PR #9, opened 2019-10-08
compute: replace template attribute with template_id

### LINKED ISSUE #8: Using IDs for template in exoscale_compute resource trigger resource re-creation when terraform apply
Hi,

Considering this code : 
```
resource "exoscale_compute" "k8s-master" {
  display_name = "ex-master-${terraform.workspace}"
  template = "${module.custom_templates.ubuntu_k8s.id}"
...
```
Will always result on a resource recreation when calling terraform apply :
```
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # module.compute_instances.exoscale_compute.k8s-master must be replaced
-/+ resource "exoscale_compute" "k8s-master" {
      ~ affinity_group_ids = [] -> (known after apply)
      ~ affinity_groups    = [] -> (known after apply)
        disk_size          = 10
        display_name       = "ex-master-default"
      ~ gateway            = "159.100.252.1" -> (known after apply)
      ~ id                 = "d9685101-c478-4d5f-a656-8eee851772fe" -> (known after apply)
        ip4                = true
        ip6                = false
      + ip6_address        = (known after apply)
      + ip6_cidr           = (known after apply)
      ~ ip_address         = "159.100.252.162" -> (known after apply)
        key_pair           = "cluster"
      ~ name               = "ex-master-default" -> (known after apply)
      ~ password           = (sensitive value)
      ~ security_group_ids = [
          - "edf52b3d-0d97-4f7f-a08d-fdf4261989b9",
          - "fb28883e-2862-4308-9394-3aa4ad329570",
        ] -> (known after apply)
        security_groups    = [
            "default",
            "k8s-master",
        ]
        size               = "Small"
      ~ state              = "Running" -> (known after apply)
      ~ tags               = {} -> (known after apply)
      ~ template           = "Linux Ubuntu 18.04 LTS Kubernetes Ready" -> "64e5af24-3859-4ff9-8f04-ed5b9331e2d2" # forces replacement
      ~ user_data_base64   = true -> (known after apply)
      ~ username           = "root" -> (known after apply)
        zone               = "ch-dk-2"

```
This issue prevent me to use my own custom template hosted at Exoscale.

Is there any workaround you know about ? Using the template name will not work as described in the documentation (using name only permit the usage of "featured" templates.)

Best regards
J.R

falzm:
This change has been prompted by issue #8, which illustrates that it's
not possible to reference an `exoscale_compute` resource `template` by
an ID (only a name) – effectively preventing the usage of custom
templates via the `exoscale_compute_template` data source.

As a consequence, the `template` attribute is now deprecated and
replaced by a new `template_id` attribute which only accepts template
IDs, which can be looked up via the `exoscale_compute_template` DS.
A collateral damage of this change is that the `username` attribute of
the resource is now also deprecated and replaced by the
`exoscale_compute_template` DS `username` attribute.

Fixes #8.

### DIFF
--- a/exoscale/datasource_exoscale_compute_template.go
+++ b/exoscale/datasource_exoscale_compute_template.go
@@ -92,22 +92,36 @@ func datasourceComputeTemplateRead(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("templates list query failed: %s", err)
 	}
 
-	templates := resp.(*egoscale.ListTemplatesResponse).Template
-	if len(templates) == 0 {
+	template := new(egoscale.Template)
+	nt := resp.(*egoscale.ListTemplatesResponse).Count
+	switch {
+	case nt == 0:
 		return errors.New("template not found")
+
+	case nt > 1:
+		return errors.New("multiple results returned, expected only one")
+
+	default:
+		template = &(resp.(*egoscale.ListTemplatesResponse).Template[0])
 	}
 
-	d.SetId(templates[0].ID.String())
+	d.SetId(template.ID.String())
 
 	if err := d.Set("id", d.Id()); err != nil {
 		return err
 	}
-	if err := d.Set("name", templates[0].Name); err != nil {
+	if err := d.Set("name", template.Name); err != nil {
 		return err
 	}
 
-	if username, ok := templates[0].Details["username"]; ok {
+	if username, ok := template.Details["username"]; ok {
 		if err := d.Set("username", username); err != nil {
+			return err
+		}
+	} else {
+		// If no username information provided in the template details,
+		// attempt an educated guess based on the template name
+		if err := d.Set("username", getSSHUsername(template.Name)); err != nil {
 			return err
 		}
 	}
--- a/exoscale/datasource_exoscale_compute_template.go
+++ b/exoscale/datasource_exoscale_compute_template.go
@@ -92,7 +92,7 @@ func datasourceComputeTemplateRead(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("templates list query failed: %s", err)
 	}
 
-	var template *egoscale.Template
+	var template egoscale.Template
 	nt := resp.(*egoscale.ListTemplatesResponse).Count
 	switch {
 	case nt == 0:
@@ -102,7 +102,7 @@ func datasourceComputeTemplateRead(d *schema.ResourceData, meta interface{}) err
 		return errors.New("multiple results returned, expected only one")
 
 	default:
-		template = &(resp.(*egoscale.ListTemplatesResponse).Template[0])
+		template = resp.(*egoscale.ListTemplatesResponse).Template[0]
 	}
 
 	d.SetId(template.ID.String())
--- a/exoscale/datasource_exoscale_compute_template.go
+++ b/exoscale/datasource_exoscale_compute_template.go
@@ -92,7 +92,7 @@ func datasourceComputeTemplateRead(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("templates list query failed: %s", err)
 	}
 
-	template := new(egoscale.Template)
+	var template *egoscale.Template
 	nt := resp.(*egoscale.ListTemplatesResponse).Count
 	switch {
 	case nt == 0:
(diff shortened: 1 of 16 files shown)


exoscale/datasource_exoscale_compute_template.go
```diff
@@ -92,24 +92,38 @@ func datasourceComputeTemplateRead(d *schema.ResourceData, meta interface{}) err
...
+	nt := resp.(*egoscale.ListTemplatesResponse).Count
+	switch {
+	case nt == 0:
 		return errors.New("template not found")
+
+	case nt > 1:
+		return errors.New("multiple results returned, expected only one")
+
+	default:
+		template = &(resp.(*egoscale.ListTemplatesResponse).Template[0])
```
pierre-emmanuelJ: Why a pointer in this case? It's prematured optimization and less readable

falzm: Fixed.

exoscale/resource_exoscale_compute.go
```diff
@@ -29,9 +29,18 @@ func resourceCompute() *schema.Resource {
 			ForceNew: true,
 		},
 		"template": {
-			Type:     schema.TypeString,
-			Required: true,
-			ForceNew: true,
+			Type:          schema.TypeString,
+			Optional:      true,
+			ForceNew:      true,
+			Deprecated:    "use `template_id` attribute with the `compute_template` data source",
```
pierre-emmanuelJ, line 35: 👍

falzm:
[ch6206]

exoscale/datasource_exoscale_compute_template.go
```diff
@@ -92,24 +92,38 @@ func datasourceComputeTemplateRead(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("templates list query failed: %s", err)
 	}
 
-	templates := resp.(*egoscale.ListTemplatesResponse).Template
-	if len(templates) == 0 {
+	template := new(egoscale.Template)
+	nt := resp.(*egoscale.ListTemplatesResponse).Count
+	switch {
```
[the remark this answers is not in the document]
pierre-emmanuelJ: My bad I thought I was looking `resource_exoscale_compute.go`.

exoscale/resource_exoscale_compute.go
```diff
@@ -225,50 +235,55 @@ func resourceComputeCreate(d *schema.ResourceData, meta interface{}) error {
...
+	_, byID := d.GetOk("template_id")
+	if !byLegacyName && !byID {
+		return errors.New("either template or template_id must be specified")
+	}
+
+	// If template is specified by name, fetch the template details from the API to get its ID.
+	// This behavior relates to the deprecated `template` attribute and will be removed in the future,
+	// where the resource will only support template IDs looked up by the compute_template data source.
+	if byLegacyName {
+		resp, err = client.RequestWithContext(ctx, &egoscale.ListTemplates{
```
pierre-emmanuelJ: For the Legacy way you can use the egoscale `GetWithContext`, that can reduce the code in this `if` statement

pierre-emmanuelJ:
For me force to put template `ID` in compute ressource It's a regression in the code, this will cause a lot of change in many users using this provider. And in Exoscale `id` of template change frequently instead of template name. I have a an example in one function that could be a solution:
https://github.com/pierre-emmanuelJ/terraform-provider-exoscale/blob/instance-pool/exoscale/resource_exoscale_instance_pool.go#L436-L443
and
https://github.com/pierre-emmanuelJ/terraform-provider-exoscale/blob/instance-pool/exoscale/resource_exoscale_instance_pool.go#L513-L537

Also this example can be used in datasource template.

In Conclusion:
it can reduce the code and increase readability and no regression, don't break the compatibility for the users using the provider,  the actual user can continue to use template name and don't care about the `ID` if It's change, custom template can be used as name easily also.

mcorbin (approved):

pierre-emmanuelJ (approved):

exoscale/resource_exoscale_compute.go
falzm: Thanks, fixed :)

PR #10, opened 2019-10-13
client: use DefaultPooledClient

greut:
By using the `DefaultPooledClient`, the provider will be more gentle with the backend (probably).

https://github.com/hashicorp/terraform/blob/master/vendor/github.com/hashicorp/go-cleanhttp/cleanhttp.go#L49-L57


falzm (approved):
Nice, thanks @greut.

pierre-emmanuelJ (approved):

PR #12, opened 2019-10-17
domain_record: fix import method

falzm:
This change fixes a bug in the import method of the `domain_record`
resource, where a record import would fail if it belonged to a DNS
domain that was not the first in the list.


falzm:
Acceptance tests passed ✅.

brutasse (approved):

pierre-emmanuelJ (approved):

PR #16, opened 2019-10-28
Add provider version to HTTP client User-Agent

falzm:
This change adds the stable release version number to the HTTP client
User-Agent. The release version value in `version/version.go` is set by the HashiCorp Terraform release pipeline during the release process.


mcorbin (approved):

pierre-emmanuelJ (approved):
LGTM

PR #21, opened 2019-11-24
Fix cloudstack config example keyword.

conet:
token is not a valid keyword and by using the example one can hit the
following error: "error when getting key of section 'cloudstack':
key 'secret' not exists".

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -49,7 +49,7 @@ for async tasks to complete. Currently, this is used during the creation of
 ```ini
 [cloudstack]
 
-endpoint = "https://api.exoscale.ch/compute"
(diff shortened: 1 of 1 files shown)


website/docs/index.html.markdown
```diff
@@ -51,7 +51,7 @@ for async tasks to complete. Currently, this is used during the creation of
 
 endpoint = "https://api.exoscale.ch/compute"
```
greut:
```suggestion
endpoint = "https://api.exoscale.com/compute"
```

falzm (approved):
Thank you @conet 👍

pierre-emmanuelJ:
Thank you @conet

PR #27, opened 2019-12-03
Replace Exoscale compute endpoint /compute by /v1

pierre-emmanuelJ:
Replace Exoscale compute endpoint /compute by /v1

### DIFF
--- a/exoscale/client.go
+++ b/exoscale/client.go
@@ -10,7 +10,7 @@ import (
 
 const defaultConfig = "cloudstack.ini"
 const defaultProfile = "cloudstack"
-const defaultComputeEndpoint = "https://api.exoscale.ch/compute"
+const defaultComputeEndpoint = "https://api.exoscale.ch/v1"
 const defaultDNSEndpoint = "https://api.exoscale.ch/dns"
 const defaultTimeout = 5 * time.Minute
 const defaultGzipUserData = true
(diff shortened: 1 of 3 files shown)


falzm (approved):

exoscale/client.go
```diff
@@ -10,7 +10,7 @@ import (
 
 const defaultConfig = "cloudstack.ini"
 const defaultProfile = "cloudstack"
-const defaultComputeEndpoint = "https://api.exoscale.ch/compute"
+const defaultComputeEndpoint = "https://api.exoscale.ch/v1"
```
falzm:
```suggestion
const defaultComputeEndpoint = "https://api.exoscale.com/v1"
```

exoscale/client.go
```diff
@@ -10,7 +10,7 @@ import (
...
 const defaultDNSEndpoint = "https://api.exoscale.ch/dns"
```
falzm:
```suggestion
const defaultDNSEndpoint = "https://api.exoscale.com/dns"
```

falzm (changes requested):
We're grown-ups now, let's switch to `.com` for the API endpoints 😉

exoscale/client.go
pierre-emmanuelJ: Nice catch

falzm (approved):

illi-j (approved):
🎂

PR #32, opened 2019-12-16
Fix getZoneByName function

pierre-emmanuelJ:
Fix the implementation of the function getZoneByName.
With the actual function if the zone name param is empty
the function return the first zone found in the Exoscale Zones list.

Fix also all Terraform resource depends on this function.


pierre-emmanuelJ:
All resources are fixed and all acceptance test pass!

falzm (approved):
LGTM, thanks @pierre-emmanuelJ 👍

illi-j (approved):
🍅

PR #66, opened 2020-08-21
healthcheck: prepare the upcoming TLS checks for managed eips

pyr:


exoscale/resource_exoscale_ipaddress.go
```diff
@@ -127,8 +137,8 @@ func resourceIPAddressCreate(d *schema.ResourceData, meta interface{}) error {
 		}
 
 		req.HealthcheckPath = d.Get("healthcheck_path").(string)
-		if req.HealthcheckMode == "http" && req.HealthcheckPath == "" {
-			return errors.New("healthcheck_path must be specified in \"http\" mode")
+		if (req.HealthcheckMode == "http" || req.HealthcheckMode == "https") && req.HealthcheckPath == "" {
+			return errors.New("healthcheck_path must be specified in \"http\" or \"htts\" mode")
```
falzm:
```suggestion
			return errors.New("healthcheck_path must be specified in \"http\" or \"https\" mode")
```

falzm:
@pyr are those new attributes editable with existing EIPs? If so, you'll also have to support state change in `resourceIPAddressUpdate()` (see https://github.com/exoscale/terraform-provider-exoscale/blob/master/exoscale/resource_exoscale_ipaddress.go#L297-L335)

falzm (changes requested):
A typo to fix and possibly missing update-related code.

falzm:
You should rebase your branch on the `master` branch which has been updated to fix the failing CI tests.

pierre-emmanuelJ (approved):
Could be nice to add a test for it ;)

pyr:
waiting for exoscale/egoscale#444

pyr:
> Could be nice to add a test for it ;)

What behavior would you add tests for in this case?

falzm:
> What behavior would you add tests for in this case?

That the new parameters are correctly updated upon configuration change (e.g. set `healthcheck_skip_tls_verify = true` and `healthcheck_skip_tls_sni = "example.net"`: https://github.com/exoscale/terraform-provider-exoscale/blob/master/exoscale/resource_exoscale_ipaddress_test.go#L55-L66

pyr:
I added tests as requested. As soon as the functionality is live they should now pass. Thanks for the help!

falzm (approved):
Needed a few fixes here and there, looks ready now 👍

PR #70, opened 2020-09-01
instance_pool: support disk_size attribute update

falzm:
This change adds support for in-place updating of the `disk_size`
attribute of the `exoscale_instance_pool` resource, instead of
re-creating the resource.


pierre-emmanuelJ (approved):
LGTM Thanks

jessicatoscani (approved):

PR #93, opened 2021-03-08
Assigne the username in the template_id case when creating an exoscale_compute resource

### LINKED ISSUE #91: Creating instance with template_id should retrieve the username from the given template not root
## The issue

When creating an `exoscale_compute` resource with a [custom template created in Exoscale](https://community.exoscale.com/documentation/compute/custom-templates/) with a custom `username` like _alice_ in the documentation, at the end of the `terraform apply` when we output the `username` we got _root_ instead of the custom `userame` : alice

- The `main.tf`

  ```hcl
  resource "exoscale_compute" "instance" {
    template_id        = "65f880d4-7b2f-11eb-9439-0242ac130002"
  }
  ```
  
- The `output.tf`

  ```
  output "username" {
    value = exoscale_compute.instance.username
  }
  ```

- The `terraform apply` output :

  ```bash
  username = "root"
  ```

## The cause 

In the `resource_exoscale_compute.go` [file](https://github.com/exoscale/terraform-provider-exoscale/blob/bd695d5e55a0ace1f0f9fe93cc0cc35cabb5b5e2/exoscale/resource_exoscale_compute.go) at the line 269, we have : 

```go
if byName {
  resp, err = client.GetWithContext(ctx, &egoscale.ListTemplates{
      ZoneID:         zone.ID,
      Name:           d.Get("template").(string),
      TemplateFilter: "featured",
   })

   if err != nil {
	return err
   }

   template := resp.(*egoscale.Template)
   templateID = template.ID.String()

   if name, ok := template.Details["username"]; username == "" && ok {
      username = name
   }

} else {
  templateID = d.Get("template_id").(string)
}

if username == "" {
  log.Printf("[INFO] %s: username not found in the template details, falling back to `root`", resourceComputeIDString(d))
  username = "root"
}
```

In the _else_ (Case using the `template_id` value) we should call the API to retrieve the `username` set in the template.

```bash
$ exo vm template ls

│b3f4fb0e-6a5d-4e10-b7b5-f6260c2a663b│ Linux Ubuntu 20.10 64-bit │ 2020-12-08T08:23:55+0000 │ ch-gva-2 │ 10 GiB │

$ exo vm template show b3f4fb0e-6a5d-4e10-b7b5-f6260c2a663b
┼───────────────┼──────────────────────────────────────┼
│   TEMPLATE    │                                      │
┼───────────────┼──────────────────────────────────────┼
│ ID            │ b3f4fb0e-6a5d-4e10-b7b5-f6260c2a663b │
│ Name          │ Linux Ubuntu 20.10 64-bit            │
│ OS Type       │ Ubuntu                               │
│ Creation Date │ 2020-12-08T08:23:55+0000             │
│ Zone          │ ch-gva-2                             │
│ Disk Size     │ 10 GiB                               │
│ Username      │ ubuntu                               │
│ Password?     │ true                                 │
│ Boot Mode     │ uefi                                 │
┼───────────────┼──────────────────────────────────────┼
```

Instead of the _"falling back to root"_

pagedidier:
This change fixes an issue where the username set in a template is not assigned when we create an `exoscale_compute` resource with the `template_id`.

Before it was falling back to the default user "root". Now it make a request to fetch the template by ID.

Fixes #91

### DIFF
--- a/exoscale/resource_exoscale_compute.go
+++ b/exoscale/resource_exoscale_compute.go
@@ -260,6 +260,9 @@ func resourceComputeCreate(d *schema.ResourceData, meta interface{}) error {
 
 	username := ""
 	templateID := ""
+
+	template:= (*egoscale.Template)(nil)
+
 	_, byName := d.GetOk("template")
 	_, byID := d.GetOk("template_id")
 	if !byName && !byID {
@@ -275,14 +278,22 @@ func resourceComputeCreate(d *schema.ResourceData, meta interface{}) error {
 		if err != nil {
 			return err
 		}
-
-		template := resp.(*egoscale.Template)
-		templateID = template.ID.String()
-		if name, ok := template.Details["username"]; username == "" && ok {
-			username = name
-		}
+		template = resp.(*egoscale.Template)
 	} else {
-		templateID = d.Get("template_id").(string)
+		id := egoscale.MustParseUUID(d.Get("template_id").(string))
+
+		resp, err = client.GetWithContext(ctx, &egoscale.ListTemplates{
+			ID:         id,
+		})
+		if err != nil {
+			return err
+		}
+		template = resp.(*egoscale.Template)
+
+	}
+	templateID = template.ID.String()
+	if name, ok := template.Details["username"]; username == "" && ok {
+		username = name
 	}
 
 	if username == "" {


falzm:
Hi,

Sorry but we won't accept this change: as explained in the issue #91, the current behavior is documented and the users are expected to use the `exoscale_compute_template` data source to retrieve the correct template username.

Issue #101, opened 2021-04-23
Changing a VM's state from "Stopped" to "Running" after provisioning it "Stopped" causes an error when trying to create a DNS record for the VM


simu:
Given the following Terraform configuration

```
terraform {
  required_version = ">= 0.14"
  required_providers {
    exoscale = {
      source  = "exoscale/exoscale"
      version = "~> 0.23"
    }
  }
}

variable "ssh_key" {
  type        = string
  description = "Name of an already present SSH Keypair"
}

variable "node_state" {
  type    = string
  default = "Stopped"
}

variable "domain" {
  type        = string
  description = "Domain name to create in Exoscale DNS"
}

data "exoscale_compute_template" "ubuntu2004" {
  zone = "ch-dk-2"
  name = "Linux Ubuntu 20.04 LTS 64-bit"
}

resource "exoscale_domain" dom {
  name = var.domain
}

resource "exoscale_compute" "vm" {
  display_name       = "test-vm"
  hostname           = "test-vm"
  key_pair           = var.ssh_key
  zone               = "ch-dk-2"
  template_id        = data.exoscale_compute_template.ubuntu2004.id
  size               = "Micro"
  disk_size          = 10
  state              = var.node_state
}

resource "exoscale_domain_record" "vm" {
  domain      = exoscale_domain.dom.id
  name        = "test-vm"
  ttl         = 60
  record_type = "A"
  content     = exoscale_compute.vm.ip_address
}
```

creating the DNS A record for the VM in "Stopped" state fails with:

```
$ export EXOSCALE_API_KEY=EXO....
$ export EXOSCALE_API_SECRET=....
$ terraform apply -var ssh_key=simon.gerber@vshn2021 -var domain=c-exo-ocp4-test.appuio-beta.ch

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # exoscale_compute.vm will be created
  + resource "exoscale_compute" "vm" {
      + affinity_group_ids = (known after apply)
      + affinity_groups    = (known after apply)
      + disk_size          = 10
      + display_name       = "test-vm"
      + gateway            = (known after apply)
      + hostname           = "test-vm"
      + id                 = (known after apply)
      + ip4                = true
      + ip6                = false
      + ip6_address        = (known after apply)
      + ip6_cidr           = (known after apply)
      + ip_address         = (known after apply)
      + key_pair           = "simon.gerber@vshn2021"
[...]

PR #102, opened 2021-04-26
compute: Recompute `ip_address` attribute on state update

simu:
This PR introduces custom diff logic to ensure a compute resource's `ip_address` attribute is correctly recomputed if the old value of the attribute is `""`. This transition occurs if the resource is initially provisioned in "Stopped" state and then booted for the first time in a subsequent `terraform apply` step.

Additionally, the PR extends the compute resource acceptance test to verify that the compute resource's `ip_address` is recomputed correctly on first boot.

Note that the custom diff logic assumes that a VM's IP address remains stable over its life time, and  only forces recomputation of the `ip_address` attribute when the old IP address attribute is the empty string and the VM's state changes from "Stopped" to "Running". Doing so minimizes the churn on resources which depend on the VM's IP address attribute such as DNS records.

The implementation makes use of the `helper/customdiff` package of the Terraform plugin SDK to implement the diffing logic.

This addresses the issue reported in #101 

See also:
* https://www.terraform.io/docs/extend/resources/customizing-differences.html
* https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/customdiff#section-documentation

### DIFF
--- a/exoscale/provider_test.go
+++ b/exoscale/provider_test.go
@@ -28,7 +28,7 @@ const (
 		    zoneid=1128bd56-b4d9-4ac6-a7b9-c715b187ce11 \
 		    name="Linux Ubuntu 20.04 LTS 64-bit"
 	*/
-	testInstanceTemplateID = "23c0622f-34cd-44c3-b995-a56d436cff85"
+	testInstanceTemplateID = "a5ddefa9-7e98-40cb-94b3-e20348b878fa"
 
 	testInstanceTypeIDTiny  = "b6cd1ff5-3a2f-4e9d-a4d1-8988c1191fe8"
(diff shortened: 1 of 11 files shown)


falzm:
Hi @simu, thank you for your contribution but we won't accept it as is, because it is too specific to your use case. However, after bringing this issue internally to my colleagues we'll find a way to fix the issue at its root (i.e. the Compute instance not being assigned an IP address until it's running), this way there won't be any workaround to implement client-side.

simu:
Hi @falzm, it's good to hear that you'll address the root cause of the issue. Do you have any idea about the time frame we can expect the issue to be fixed?

falzm:
> Do you have any idea about the time frame we can expect the issue to be fixed?

Not at the moment, but we will follow-up via this issue when it's done.

PR #132, opened 2021-12-03
SKS

falzm:
## sks_cluster: change OIDC required_claim attribute type

This change updates the `exoscale_sks_cluster.oidc.required_claim`
attribute to be a map type instead of a string, following recent
Exoscale API change.

## sks_nodepool: add support for K8s taints

This change adds support for Kubernetes taints to the
`exoscale_sks_nodepool` resource.


falzm:
[sc-38636]

PR #133, opened 2021-12-06
Switch exoscale_security_group* resources to public API

falzm:
This changes overhauls the `exoscale_security_group*` resources and
`exoscale_security_group` data source to use the new Exoscale public API
(aka V2).

### DIFF
--- a/website/docs/r/security_group.html.markdown
+++ b/website/docs/r/security_group.html.markdown
@@ -15,8 +15,9 @@ Provides an Exoscale [Security Group][sg-doc] resource. This can be used to crea
 
 ```hcl
 resource "exoscale_security_group" "web" {
-  name        = "web"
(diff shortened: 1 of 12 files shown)


website/docs/r/security_group.html.markdown
```diff
@@ -27,6 +28,7 @@ In addition to the arguments listed above, the following attributes are exported
 
 * `name` - (Required) The name of the Security Group.
 * `description` - A free-form text describing the Anti-Affinity Group purpose.
```
exo-cedric:
```suggestion
* `description` - A free-form text describing the Security Group purpose.
```

falzm: Good catch, thanks.

exo-cedric (commented):
Thx!

PR #146, opened 2022-02-01
sks_cluster: export kubeconfig string as a resource attribute

### LINKED ISSUE #131: feature-request: sks: export kubeconfig string (or even separate credentials) of the resource
Hello!
I'd really like to have ability not just create SKS cluster using terraform, but at the same time get the connection credentials to use it in-place with kubernetes/helm terraform providers.
It is kind of enabler for me to use SKS on Exoscale.
As I can understand there is pretty straight forward possibility to implement such feature.
However if there is any cons which prevents of doing so please let me know.

PS: most of managed kubernetes providers like DO, GCP, AWS, YACloud have this ability.

Strangerxxx:
should close #131

---

simply gets the kubeconfig string from the exoscale api and sets it to the resource attribute

### DIFF
--- a/exoscale/resource_exoscale_sks_cluster.go
+++ b/exoscale/resource_exoscale_sks_cluster.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"time"
 
 	egoscale "github.com/exoscale/egoscale/v2"
 	exoapi "github.com/exoscale/egoscale/v2/api"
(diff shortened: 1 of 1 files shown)


Strangerxxx:
i'm not 100% sure, that I did it right as of project code style stand point
any comments will be appreciated

dramaturg:
We have so far been reluctant to add something like this because we were concerned about storing secrets in Terraform state. However, there were some requests to have something like this a couple of times and it would make sense to accommodate that.

Your approach is straightforward but it has some drawbacks:
* Our API generates a new kubeconfig on every request. That means without some extra care, it'll change on every single Terraform run and it would do so also for users who aren't using this.
* The kubeconfig secret will be in Terraform state and there would be no way to avoid that.

After some discussion we came to the conclusion that having our provider support exporting kubeconfig would have to:
* implement a new resource such as `exoscale_sks_cluster_kubeconfig` so every user can decide if they want to accept having a `system:masters` kubeconfig stored in Terraform state
* the implementation would have to parse the exported kubeconfig and check the contained certificate to only re-request a new config from the API once it expires

Of course all that is significantly more work than your approach so it's perfectly understandable if you don't want to go there. We'll keep it on our roadmap in any case but we can't say if we will get around to it any time soon.

Kind regards from Exoscale!

PhilippeChepy:
This solution is implemented in #159 we are going to release soon.
Therefore, we can close this PR.

PR #159, opened 2022-03-08
Add exoscale_sks_kubeconfig resource

PhilippeChepy:
Implement a solution for https://github.com/exoscale/terraform-provider-exoscale/issues/131.

We decided to implement the Kubeconfig generation in a separate resource to let every SKS user the choice to use it or not:
- If you want to generate Kubeconfigs from Terraform, it gives lots of flexibility, allowing you to build multiple configuration embedding certificates with fine-grained control on the user (CN), groups (orgs), or TTL of the client certificate.
- If you want to keep secrets outside of your Terraform state, just continue using the provider as usual, without declaring any `exoscale_sks_kubeconfig`.

### DIFF
--- a/docs/resources/sks_kubeconfig.md
+++ b/docs/resources/sks_kubeconfig.md
@@ -49,8 +49,8 @@ output "kubeconfig" {
 * `cluster_id` - (Required) The ID of the target SKS cluster.
 * `ttl_seconds` - The number of seconds after initial issuing that the Kubeconfig will become invalid.
 * `early_renewal_seconds` - If set, the resource will consider the Kubeconfig to have expired the given number of seconds before its actual ca certificate or client certificate expiry time. This can be useful to deploy an updated Kubeconfig in advance of the expiration of its internal current certificate. Note however that the old certificate remains valid until its true expiration time since this resource does not (and cannot) support certificate revocation. Note also that this advance update can only be performed should the Terraform configuration be applied during the early renewal period.
-* `user` - User name in the generated Kubeconfig. The certificate present in the Kubeconfig will also have this name set for the CN field.
-* `groups` - Group names in the generated Kubeconfig. The certificate present in the Kubeconfig will have these roles set in the Organization field.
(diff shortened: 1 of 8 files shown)


dramaturg (approved):

docs/resources/sks_kubeconfig.md
```diff
@@ -0,0 +1,70 @@
...
+  sensitive = true
+}
+```
+
+
+## Arguments Reference
+
+* `zone` - (Required) The name of the [zone][zone] of the target SKS cluster.
+* `cluster_id` - (Required) The ID of the target SKS cluster.
+* `ttl_seconds` - The number of seconds after initial issuing that the Kubeconfig will become invalid.
```
exo-cedric:
```suggestion
* `ttl_seconds` - The number of seconds after initial issuing that the Kubeconfig will become invalid (default: 2592000 = 30 days).
```

docs/resources/sks_kubeconfig.md
```diff
@@ -0,0 +1,70 @@
...
+* `early_renewal_seconds` - If set, the resource will consider the Kubeconfig to have expired the given number of seconds before its actual ca certificate or client certificate expiry time. This can be useful to deploy an updated Kubeconfig in advance of the expiration of its internal current certificate. Note however that the old certificate remains valid until its true expiration time since this resource does not (and cannot) support certificate revocation. Note also that this advance update can only be performed should the Terraform configuration be applied during the early renewal period.
```
exo-cedric:
```suggestion
* `early_renewal_seconds` - If set, the resource will consider the Kubeconfig to have expired the given number of seconds before its actual ca certificate or client certificate expiry time. This can be useful to deploy an updated Kubeconfig in advance of the expiration of its internal current certificate. Note however that the old certificate remains valid until its true expiration time since this resource does not (and cannot) support certificate revocation. Note also that this advance update can only be performed should the Terraform configuration be applied during the early renewal period (default: 0).
```

exo-cedric (approved):

PR #171, opened 2022-04-21
fix infinite version attribute update for PostgreSQL DBaaS

Strangerxxx:
The issue is in that the version attribute is set from terraform module must contain only major number.
However is not the same as on api backend where it is representing actual version containing minor number as well.

This changes introduces actual_version attribute to pg section of exoscale_database resource to save an actual version (including minor number) which comes from api and compare it with the version attribute with semver comparison.


exoscale/resource_exoscale_database_pg.go
```diff
@@ -414,7 +423,12 @@ func resourceDatabaseApplyPg(ctx context.Context, d *schema.ResourceData, client
...
 
 	if v := databaseService.Version; v != nil {
-		pg[resDatabaseAttrPgVersion] = *v
+		pg[resDatabaseAttrPgActualVersion] = *v
+	}
+
+	if v := databaseService.Version; v != nil {
+		version, _ := semver.NewVersion(*v)
+		pg[resDatabaseAttrPgVersion] = strconv.FormatUint(version.Major(), 10)
 	}
```
7fELF: I would suggest this instead, this is simple enough and avoids an additional dependency.

```suggestion
	if databaseService.Version != nil {
		pg[resDatabaseAttrPgVersion] = strings.SplitN(*databaseService.Version, ".", 2)[0]
	}
```

Also, do you think `actual_version` is useful? Would you use it?

Strangerxxx: I agree with your suggestion regarding simplification and avoiding redundant dependency.
However I do think that `actual_version` may be useful in case of tracking changes applied externally especially from the infosec perspective.
It would be more useful if I could set exact version indeed, but it is another story I guess :)

hbollon:
Any updates on this?

kobajagi:
Fixed in https://github.com/exoscale/terraform-provider-exoscale/commit/e7d4b6240757af6647a8053a8e7e24cc74d1566c
For now without `actual_version` attribute.

Thanks @Strangerxxx  for contribution, it wasn't merged but it provided useful insight.

PR #172, opened 2022-04-21
add parsed attributes for sks_kubeconfig resource

Strangerxxx:
* refactor to use full-featured kubeconfig parser using clientcmd package
* refactor certificates parsing and ID calculation
* update tests


PhilippeChepy:
Hello @Strangerxxx,

Thank you for this contribution!

While storing Kubeconfig attributes in a split way sounds good from a semantic point of view, we want to keep the dependencies of our Terraform provider as light as possible. As the Kubernetes go client SDK is *heavy* and we only need a subset of its features, we would like to _stick with the current Kubeconfig parsing code which is way lighter_ (it only relies on the YAML parsing library as a unique dependency).
As your implementation is also a breaking change, we also need to _take care of the state migration_.

We will implement the new feature from your contribution ourselves (no ETA), taking care of the abovementioned criteria.

Cheers

Strangerxxx:
Thank you for response on my PR, @PhilippeChepy!

I agree with all points you raised here.
I will try to do it by myself and update this PR with all the optimizations you've proposed.
I will revert it back to the simple yaml parsing with less dependencies and will try to implement state migration for it.

However, at this point the only thing which still struggling me is that cluster CA certificate is still can be exported from kubeconfig resource only.
It is a problem for me, because requires me to apply with a target on kubeconfig resource first in order to reissue it before I can use it with terraform-kubernetes-provider configuration (with credentials plugin).
As I understand in general CA certificate is not expiring in the way of Client certificate do, however it still requires me to do so.

As @jessicatoscani posted in #131
> The certificate-authority-data content of kubeconfig is the apiserver CA. The command exo compute sks authority-cert can only provide [kubelet or aggregation authorities](https://community.exoscale.com/documentation/sks/certificates-and-keys/#certificates-authorities)

As I can understand it is the [API limitation](https://openapi-v2.exoscale.com/#operation-get-sks-cluster-authority-cert), where for some reason there is no option to get the apiserver CA.

If there will be an option to have CA certificate with the sks_cluster resource itself, so there is no actual need to have those attributes separately in the sks_kubeconfig resource (or even have it at all).

Is there any plans to extend the API with ability to get cluster CA and have it along with sks_cluster itself?

PhilippeChepy:
So, for your use case, a good solution could be to have either (or both): 
- a data source (something like `sks_ca_certificates`) to retrieve the cluster's CA certificates (at least the client CA).
- export these CA certificates as attributes straight from the `sks_cluster` resource.

You're right, for now, the API currently doesn't export the client CA certificate. This would be a prerequisite to implementing the abovementioned solutions and will be discussed internally.

PhilippeChepy:
Hello @Strangerxxx,

The API has been updated lately, with the ability to export the control-plane CA certificate.

We will start by providing exported attributes on the `sks_cluster` resource, removing the need for this PR.
A PR is landing for this update (#201), with documentation and tests.

It relies on our API to retrieve:
- Control plane CA certificate (the one you need)
- Kubelet CA certificate
- Aggregation layer CA certificate

Again, thank you for your initial contribution and for pointing out this issue.

PR #185, opened 2022-06-22
Get rid of any specific template ID from acceptance tests

PhilippeChepy:


exoscale/datasource_exoscale_compute_template_test.go
```diff
@@ -18,8 +17,6 @@ var (
 )
 
 func TestAccDataSourceComputeTemplate(t *testing.T) {
-	templateID := "4a850c9d-93f4-4b39-b4d7-2cbf3a1f1227"
```
PhilippeChepy, line 21: Here is the id we had to update continuously.

exoscale/provider_test.go
```diff
@@ -24,16 +24,6 @@ const (
...
 
-	/*
-		Reference template used for tests: "Linux Ubuntu 20.04 LTS 64-bit" @ de-muc-1 (featured)
-
-		cs --region cloudstack listTemplates \
-		    templatefilter=featured \
-		    zoneid=85664334-0fd5-47bd-94a1-b4f40b1d2eb7 \
-		    name="Linux Ubuntu 20.04 LTS 64-bit"
-	*/
-	testInstanceTemplateID = "4a850c9d-93f4-4b39-b4d7-2cbf3a1f1227"
```
PhilippeChepy, line 35: The same ID was also here.

exo-cedric (approved):
:100:

arapaho (approved):

PR #227, opened 2022-11-07
Implement labels on Elastic IPs

PhilippeChepy:
Implement labels on Elastic IP resource and datasource.

```
➜ TF_ACC=true go test -timeout 600s -run ^TestAccDataSourceElasticIP$ github.com/exoscale/terraform-provider-exoscale/exoscale
ok      github.com/exoscale/terraform-provider-exoscale/exoscale        17.130s
```

```
➜ TF_ACC=true go test -timeout 600s -run ^TestAccResourceElasticIP$ github.com/exoscale/terraform-provider-exoscale/exoscale
ok      github.com/exoscale/terraform-provider-exoscale/exoscale        35.596s
```


PhilippeChepy:
This branch needs https://github.com/exoscale/egoscale/pull/566 to be merged first

exoscale/resource_exoscale_elastic_ip.go
```diff
@@ -213,6 +219,14 @@ func resourceElasticIPCreate(ctx context.Context, d *schema.ResourceData, meta i
 		}
 
 		elasticIP.Healthcheck = &elasticIPHealthcheck
+
+		if l, ok := d.GetOk(resElasticIPAttrLabels); ok {
+			labels := make(map[string]string)
+			for k, v := range l.(map[string]interface{}) {
```
kobajagi, line 225: Casting to `map[string]string` should work.

exoscale/resource_exoscale_elastic_ip.go
```diff
@@ -279,6 +293,15 @@ func resourceElasticIPUpdate(ctx context.Context, d *schema.ResourceData, meta i
 
 	var updated bool
 
+	if d.HasChange(resElasticIPAttrLabels) {
+		labels := make(map[string]string)
+		for k, v := range d.Get(resElasticIPAttrLabels).(map[string]interface{}) {
+			labels[k] = v.(string)
+		}
+		elasticIP.Labels = &labels
```
kobajagi, line 301: I don't see the need to copy labels by value. Assuming my previous suggestion works (casting to `map[string]string`), we can just set pointer to that. I could be wrong though, to be tested.

kobajagi: Same snippet is all over our codebase, but if simplification works we might as well start code cleanup from this.

kobajagi (approved):
Great work!

exoscale/resource_exoscale_elastic_ip.go
PhilippeChepy: It should (intuitively), but when trying to cast  `l` directly into a `map[string]string` it leads to:

```
panic: interface conversion: interface {} is map[string]interface {}, not map[string]string
```

kobajagi: I guess i was wrong. Thanks for checking.

jessicatoscani (approved):

PR #228, opened 2022-11-08
FIxing acceptance tests

kobajagi:
Fixes issues with acceptance tests. List of changes:
- terraform SDK updated,
- disabled `MaxIndexCount` check for opensearch database resource (upstream bug),
- kafka version in tests updated,
- multiple import tests fixed.


exoscale/resource_exoscale_database_kafka_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var (
 	testAccResourceDatabaseKafkaIPFilter = []string{"1.2.3.4/32"}
-	testAccResourceDatabaseKafkaVersion  = "3.1"
+	testAccResourceDatabaseKafkaVersion  = "3.2"
 	testAccResourceDatabasePlanKafka     = "business-4"
```
PhilippeChepy, line 18: It could be nice if we can fix this value durably (maybe to be handled in a separate story).

PhilippeChepy (approved):
:rocket:

exoscale/resource_exoscale_database_kafka_test.go
kobajagi: I agree. Same applies for other databases so I'll create a separate story.

kobajagi: Story created.

jessicatoscani (approved):

PR #239, opened 2023-01-23
Remove nested imports from resources domain and security group

kobajagi:
This PR removes nested import feature from `exoscale_security_group` and `exoscale_domain` resources.
Nested import no longer works with terraform 1.3.2 and above. As feature is not straightforward to use anyhow (due to generated names for nested resources) i want to remove it instead of trying to fix it. Nested resources should be imported individually.


PhilippeChepy (approved):
LGTM :+1: 

Did you try to import a security-group rule and a domain record independently?
It seems we have the code for that, but I never tried.

kobajagi:
@PhilippeChepy Yes, tried and works fine. Import functionality of `domain_record` and `security_group_rule` is already covered by acc tests.

jessicatoscani (approved):

exo-cedric (approved):

PR #245, opened 2023-02-16
Add data sources for SKS clusters and nodepools

sauterp:
We add four new data sources `exoscale_sks_cluster`, `exoscale_sks_cluster_list`, `exoscale_sks_nodepool` and `exoscale_sks_nodepool_list`. We also add the new package `pkg/gen/datasource` which generalizes the creation of filtered list data sources. A function reads the schema of a singular data source and returns a list data source that can filter for attributes in the input data source. The developer only needs to implement the interactions with the exoscale API by hand.


pkg/gen/datasource/datasource.go
```diff
@@ -0,0 +1,21 @@
...
+	if value == nil {
+		return
+	}
+
+	data[attributeIdentifier] = value.Format(time.RFC3339)
+}
+
+func Assign[T any](data TerraformObject, attributeIdentifier string, value *T) {
+	if value == nil {
+		return
```
PhilippeChepy: If these helpers can also be used for resources, we should choose another name than "datasource".

PhilippeChepy (commented):
Nice!
Don't forget to update the documentation :)

jessicatoscani (approved):

exoscale/datasource_exoscale_sks_nodepool.go
```diff
@@ -0,0 +1,130 @@
...
+		return diag.Errorf("error getting cluster %q: %s", clusterID, err)
+	}
+
+	filters, err := filter.CreateFilters(ctx, d, dataSourceSKSNodepool().Schema)
+	if err != nil {
+		return diag.Errorf("failed to create filter: %q", err)
+	}
+
+	var matchingNodePool datasource.TerraformObject
+	nMatches := 0
```
kobajagi, line 102: I don't see where `nMatches` will change value, it is always 0?

newbenji:
Really would love this one to be approved

newbenji:
Would it be possible it can get and kubeconfig also from this?

PhilippeChepy:
> Would it be possible it can get and kubeconfig also from this?

Hello @newbenji, you can get a kubeconfig file thanks to the [exoscale_sks_kubeconfig](https://registry.terraform.io/providers/exoscale/exoscale/latest/docs/resources/sks_kubeconfig) resource.

We are not willing to implement kubeconfig generation as a datasource because it's not something we store on the api side, and it has it's own lifecycle (e.g. because of its TTL).

newbenji:
ah then just hope we soon have a new version with this pullrequest in

PhilippeChepy (approved):

PR #257, opened 2023-04-17
fixed typo in nlb_service description

horakg:
fixed some typo in the nlb_service description

### DIFF
--- a/docs/resources/nlb_service.md
+++ b/docs/resources/nlb_service.md
@@ -74,7 +74,7 @@ directory for complete configuration examples.
 
 Required:
 
-- `port` (Number) The NLB service (TCP/UPP) port.
+- `port` (Number) The NLB service (TCP/UDP) port.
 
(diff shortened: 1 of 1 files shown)


docs/resources/nlb_service.md
```diff
@@ -74,7 +74,7 @@ directory for complete configuration examples.
 
 Required:
 
-- `port` (Number) The NLB service (TCP/UPP) port.
+- `port` (Number) The NLB service (TCP/UDP) port.
```
jessicatoscani, line 77: This documentation is generated from code, it should be fixed in the resource itself, here:
https://github.com/exoscale/terraform-provider-exoscale/blob/4315c0221adfb27094f7256fd467f38f06327c2e/exoscale/resource_exoscale_nlb_service.go#L78

horakg: will do

PR #268, opened 2023-05-22
Make `compute_instance.network_interface.ip_address` a computed field

### LINKED ISSUE #264: exoscale_compute_instance network_interface always updated
I want exoscale to assign my instance an ip address via dhcp

I don't want to hardcode the ip_address into my terraform

If I don't declare ip_address within network_interface then on every terraform run it wants to update the resource even though nothing is changing

This used to work in exoscale_compute, how do I get the same behaviour with exoscale_compute_instance?

kobajagi:
This PR fixes #264

### DIFF
--- a/pkg/resources/instance/resource.go
+++ b/pkg/resources/instance/resource.go
@@ -94,6 +94,7 @@ func Resource() *schema.Resource {
 						Description:      "The IPv4 address to request as static DHCP lease if the network interface is attached to a *managed* private network.",
 						Type:             schema.TypeString,
 						Optional:         true,
+						Computed:         true,
 						ValidateDiagFunc: validation.ToDiagFunc(validation.IsIPv4Address),
 					},
 					"network_id": {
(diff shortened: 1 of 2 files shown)


sauterp (approved):

sauterp:
Maybe I'm misunderstanding the intended behaviour of this resource, but I think this produces another problem. Now that the "ip_address" attribute is `computed`, it means that the user cannot set this field anymore manually. But AFAIK he should have the option to do that.

kobajagi:
> Maybe I'm misunderstanding the intended behaviour of this resource, but I think this produces another problem. Now that the "ip_address" attribute is `computed`, it means that the user cannot set this field anymore manually. But AFAIK he should have the option to do that.

It should still allow you to set the value manually. I could add the specific test case to avoid confusion.

kobajagi:
Added test to check managed private network with defined `ip_address`.

PhilippeChepy (approved):
Thank you!

PR #270, opened 2023-05-31
zones data source

sauterp:
We add a data source for zones. This is the first data source we implemented in the terraform-plugin-framework. We use a mux to combine the SDK provider with the new Framework provider. Some refactoring was necessary, since the two providers need to provide the same schema and configuration.


examples/zones/README.md
```diff
@@ -0,0 +1 @@
+# Zones
```
PhilippeChepy, line 1: Is this file expected to contains only the title?

PhilippeChepy (approved):
lgtm

Thank you for introducing the new framework :)

examples/zones/README.md
sauterp: sorry I forgot to write it^^

kobajagi (approved):

PR #271, opened 2023-06-06
docs: wrap ! symbol explanation in a box 

kobajagi:
Updated resource template to separate ! symbol note from timeout schema block (check [here](https://registry.terraform.io/providers/exoscale/exoscale/0.48.0/docs/resources/compute_instance) how it looks now).


sauterp (approved):
good catch! :sweat_smile:

jessicatoscani (approved):

PR #278, opened 2023-07-11
Disable user_data gzip compression

PhilippeChepy:
Disable gzip of user_data, since it could be handled by the [`cloudinit` datasource](https://registry.terraform.io/providers/hashicorp/cloudinit/latest/docs/data-sources/config)

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -29,7 +29,6 @@ const (
 	DnsEndpointAttrName     = "dns_endpoint"
 	EnvironmentAttrName     = "environment"
 	TimeoutAttrName         = "timeout"
-	GZIPUserDataAttrName    = "gzip_user_data"
 	DelayAttrName           = "delay"
 )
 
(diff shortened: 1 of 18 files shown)


pkg/provider/provider.go
```diff
@@ -101,12 +99,6 @@ func (p *ExoscaleProvider) Schema(ctx context.Context, req provider.SchemaReques
 					"Timeout in seconds for waiting on compute resources to become available (by default: %.0f)",
 					config.DefaultTimeout.Seconds()),
 			},
-			GZIPUserDataAttrName: schema.BoolAttribute{
-				Optional: true,
-				MarkdownDescription: fmt.Sprintf(
-					"Defines if the user-data of compute instances should be gzipped (by default: %t)",
-					config.DefaultGzipUserData),
-			},
```
kobajagi, line 109: Perhaps we should keep the attribute but make it `no-op` and deprecated instead? Just so it does not break any existing configuration.

PhilippeChepy: This attribute hasn't any impact on the provider behavior even with currently released versions.

I'm more in favor of removing it completely since any broken configuration could be fixed by just removing this optional attribute.

kobajagi: Lets do that and add note about in `CHANGELOG.md`.

sauterp (approved):

kobajagi (approved):

Issue #272, opened 2023-06-19
Feature Request: data source for nlb services


maxlareo:
Hi,

It'd be nice to be able to retrieve data for NLB Services not managed by Terraform.

My use case is the following: I have an SKS cluster using the Exoscale cloud controller, Kubernetes ingress-controller is creating an Exoscale NLB with NLB services automatically and now I would like to setup security groups regarding the NLB services.

I currently don't see a way to retrieve NLB services ports and HC ports numbers from Exoscale Terraform provider in this use case, the workaround I see is to use exoscale_sks_kubeconfig and then use Kubernetes provider to retrieve the relevant data from Kubernetes resources.

This is not ideal as I'd like to avoid resorting to store a kubeconfig in the Terraform state, even one linked to a read-only group on restricted resources.

What do you think ?

maxlareo:
I could try implementing that feature if you agree with it.

Just tell me what do you prefer, from what I see we can either:
1. Return the list of services in the data source `exoscale_nlb`.
2. Add a dedicated data source `exoscale_nlb_services` that return the list of service for a given NLB ID or Name.
3. Add a dedicated data source `exoscale_nlb_service` that return a single service for a given NLB ID or Name and a given NLB Service ID or Name.

kobajagi:
Hello @maxlareo 

Thank you for your interest in the exoscale terraform-provider. Data source for NLB services is on our roadmap but we don't have any ETA. We will gladly provide guidance if you are interested in implementing the feature yourself, however keep in mind that we are currently migrating our codebase to the new terraform plugin development library ([terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework)) and new data source will need to use framework (check [zones](https://github.com/exoscale/terraform-provider-exoscale/blob/master/pkg/resources/zones/datasource.go) data source as an example).

To be inline with our codebase it should be a dedicated data source `exoscale_nlb_service_list` that returns the list of NLB services (second option you suggested). You can get the list of services in [this client call](https://pkg.go.dev/github.com/exoscale/egoscale/v2#Client.FindNetworkLoadBalancer).

Let us know if you need help at any point.

maxlareo:
The datasource is now available on the latest version released on Terraform.

I can ditch the Kubernetes provider usage and stop relying on a kubeconfig stored on Terraform state file. This feels much more cleaner.

Thanks !

PR #282, opened 2023-07-24
Add nlb_service_list datasource

maxlareo:
Related to #272.

This is an attempt to add the datasource `exoscale_nlb_service_list` that would allow to retrieve given Exoscale NLB services information as described in the related issue.

### DIFF
--- a/pkg/resources/nlb_service/datasource_list.go
+++ b/pkg/resources/nlb_service/datasource_list.go
@@ -217,6 +217,7 @@ func (d *NLBServiceListDataSource) Read(ctx context.Context, req datasource.Read
 	}
 
 	data.NLBName = types.StringValue(*nlb.Name)
+	data.NLBID = types.StringValue(*nlb.ID)
 
 	for _, service := range nlb.Services {
 		var serviceState Service
--- a/pkg/resources/nlb_service/datasource_list.go
+++ b/pkg/resources/nlb_service/datasource_list.go
@@ -216,8 +216,12 @@ func (d *NLBServiceListDataSource) Read(ctx context.Context, req datasource.Read
 		return
 	}
 
-	data.NLBName = types.StringValue(*nlb.Name)
-	data.NLBID = types.StringValue(*nlb.ID)
+	if nlb.ID != nil && data.NLBID.IsNull() {
+		data.NLBID = types.StringValue(*nlb.ID)
+	}
+	if nlb.Name != nil && data.NLBName.IsNull() {
+		data.NLBName = types.StringValue(*nlb.Name)
+	}
 
 	for _, service := range nlb.Services {
 		var serviceState Service
--- a/pkg/resources/nlb_service/datasource_list.go
+++ b/pkg/resources/nlb_service/datasource_list.go
@@ -1,0 +1,289 @@
+package nlb_service
+
+import (
+	"context"
+	"time"
+
+	exoscale "github.com/exoscale/egoscale/v2"
+	exoapi "github.com/exoscale/egoscale/v2/api"
+
+	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+
+	"github.com/exoscale/terraform-provider-exoscale/pkg/config"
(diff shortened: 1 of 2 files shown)


maxlareo:
Hi @kobajagi,

I'm trying to implement the datasource as discussed, this is my first time getting my hands into a terraform provider, could you check if I'm starting properly with the file architecture and the datasource structure as well please ?

To ensure I'm good before starting to actually work on the logic of the datasource via the Read function.

kobajagi:
Hey @maxlareo. Thank you very much for this PR. You are on the right track, just a few small corrections:
- rename folder `pkg/resources/nlb_service_list` to `pkg/resources/nlb_service` to be consistent with our naming (eg. `instance_pool_list` datasouce code is within `pkg/resources/instance_pool`;
- you can also drop the full name from filenames and symbols (eg. `datasource_exoscale_nlb_service_list.go` to `datasouce_list.go` and `NLBServiceHealthcheckDataSourceModel` to `DataSourceListModel`) because package name already contains resource name;
- add timeout  (check [here](https://github.com/exoscale/terraform-provider-exoscale/blob/c45eaf0db30cfcf16ecb579b7c030c622ae86a80/pkg/resources/database/datasource_uri.go#L45) for hints).

Feel free to ping me here if you need assistance.

maxlareo:
I have a working version, it's pretty naive implementation but I manage to get all services data from a given NLB ID or Name.

If you could have an eye and tell me if that's ok or what needs to be improved, thanks!

pkg/resources/nlb_service/datasource_list.go
```diff
@@ -0,0 +1,289 @@
...
+			healtcheckState.URI = types.StringValue(*service.Healthcheck.URI)
+		}
+
+		serviceState.Healthcheck = healtcheckState
+
+		data.NLBServiceList = append(data.NLBServiceList, serviceState)
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
```
kobajagi: Not needed as this is the last statement anyway.

kobajagi (approved):

kobajagi:
@maxlareo All good, just a minor suggestion. Thanks!
I'll write some tests to cover the PR and then it can be merged.

pkg/resources/nlb_service/datasource_list.go
```diff
@@ -0,0 +1,289 @@
...
+			err.Error(),
+		)
+		return
+	}
+
+	data.NLBName = types.StringValue(*nlb.Name)
+
+	for _, service := range nlb.Services {
+		ctx = tflog.SetField(ctx, "exoscale_service", service)
+		tflog.Info(ctx, "Loop NLB service")
```
sauterp: It seems to me this line doesn't provide much useful information, can it be removed?

maxlareo: Sure, I'll remove these log related lines.

pkg/resources/nlb_service/datasource_list.go
```diff
@@ -0,0 +1,285 @@
...
+	nlb, err := d.client.FindNetworkLoadBalancer(ctx, data.Zone.ValueString(), x)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Unable to find Network Load Balancer",
+			err.Error(),
+		)
+		return
+	}
+
+	data.NLBName = types.StringValue(*nlb.Name)
```
maxlareo: I just saw that I forgot to assign the NLB ID in case the NLB Name is given, this will result in empty NLB ID field from the state. I'm adding the missing attribution.

pkg/resources/nlb_service/datasource_list.go
maxlareo: That's what Terraform tutorial shows: https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-data-source-read#implement-read-functionality

Is it not possible the line above might generate an error which wouldn't be raised without it ?

kobajagi: I don't see how it would make any difference, on error it just calls `return` to interrupt the flow but there is no code after it that would execute. Tutorial probably expects you would extend it and add more code after it.

Here is an example from hashicorp themselves that doesn't check for error on the last line:
https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/internal/provider/example_data_source.go#L104

maxlareo: Thanks for the confirmation, I'll remove it.

pkg/resources/nlb_service/datasource_list.go
sauterp: I think it would also be good to check if nlb.Name and nlb.ID are nil before dereferencing them.

kobajagi:
I've rebased the branch, working on the tests right now.

pkg/resources/nlb_service/datasource_list.go
maxlareo: I've added checks regarding this on https://github.com/exoscale/terraform-provider-exoscale/pull/282/commits/401c33ad69b87ef7f433afc70a5a7281b5c06448

PhilippeChepy (approved):

PR #288, opened 2023-07-28
Prevent // run of acc-tests

jessicatoscani:


.github/workflows/tests.yml
```diff
@@ -11,8 +12,15 @@ on:
...
 
+# Prevent // execution of acc tests
+concurrency: acc-tests
+
 jobs:
   test:
+    # if test job is already running for the same branch, cancel it and run this one
+    concurrency:
+      group: acc-tests-${{ github.ref }}
+      cancel-in-progress: true
```
jessicatoscani: I have tested this one, as we it will kill a running job in the middle of the execution, it might leave orphaned resources in the test org. It could be remove later if there are more disadvantages than advantages

kobajagi: Shouldn't we instead queue tests to run in one thread (wait for running one to finish before starting next)?

jessicatoscani: Small precision, if you work on another branch, tests will be queued and not cancelled. The cancellation will only happen for your own branch.

If you push new changes in your branch, I think you would prefer to run the tests on this last commit rather than an old one. 
Worst case, you can skip the execution of the workflow with ` [skip ci]` ([ref](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs).

jessicatoscani: Only queue jobs: https://github.com/exoscale/terraform-provider-exoscale/pull/288/commits/93230e6859bce795d3091f3fd492fe5bbbe925ee

kobajagi (approved):

yannprvst (approved):

PR #337, opened 2024-02-20
compute_instance: add `destroy_protected` attr

sauterp:
# Description
We add the attribute `destroy_protected` to the `exoscale_compute_instance` resource. By default it is set to false. If the attribute is set to true the API will return an error, if an attempt is made to delete the affected instance.

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing
Two new acceptance tests have been added that verify the feature. The failing acceptance tests in CI are not related to this feature in any way.

### DIFF
--- a/pkg/resources/instance/resource.go
+++ b/pkg/resources/instance/resource.go
@@ -38,7 +38,6 @@ func Resource() *schema.Resource {
 			Description: "Mark the instance as protected, the Exoscale API will refuse to delete the instance until the protection is removed (boolean; default: `false`).",
 			Type:        schema.TypeBool,
 			Optional:    true,
-			Computed:    true,
 		},
 		AttrDeployTargetID: {
 			Description: "A deploy target ID.",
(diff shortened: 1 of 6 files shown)


pierre-emmanuelJ (approved):
Nice one 🎉 
Just missing the acceptance test to be OK, and we are good!

pkg/resources/instance/resource.go
```diff
@@ -34,6 +34,12 @@ func Resource() *schema.Resource {
 			Type:        schema.TypeString,
 			Computed:    true,
 		},
+		AttrDestroyProtected: {
+			Description: "Mark the instance as protected, the Exoscale API will refuse to delete the instance until the protection is removed (boolean; default: `false`).",
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Computed:    true,
```
kobajagi: It should not be `Computed` (just `Optional`) as we don't read remote value.

sauterp: Good point. I made it `Computed` because in the long run we plan to implement API calls that allow us to read the remote value. But now I removed it and tested that it can be added later without invalidating an existing and applied tf config. So TIL: adding the `Computed` property to an attribute is a backwards compatible change from the configuration point of view.

However, if you read the comment in my latest commit, you'll see that this brings a slight change in behaviour. But I think this is a good solution.

kobajagi (approved):

PR #341, opened 2024-04-04
Block storage volume resource & data source

kobajagi:
# Description
This PR adds `exoscale_block_storage_volume` resource & data source.
Also extends `exoscale_compute_instance` resource to allow attaching/detaching volumes (tests are in the `block_storage` package).

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK


docs/data-sources/block_storage_volume.md
```diff
@@ -0,0 +1,67 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "exoscale_block_storage_volume Data Source - terraform-provider-exoscale"
+subcategory: ""
+description: |-
+  Fetch Exoscale Block Storage https://community.exoscale.com/documentation/block-storage/ Volume.
+  Block Storage offers persistent externally attached volumes for your workloads.
+  Corresponding resource: exoscaleblockstorage_volume ../resources/block_storage_volume.md.
```
pierre-emmanuelJ, line 8:
```suggestion
  Corresponding resource: exoscale_block_storage_volume ../resources/block_storage_volume.md.
```

pkg/resources/block_storage/resource_volume.go
```diff
@@ -0,0 +1,588 @@
...
+			},
+			"name": schema.StringAttribute{
+				MarkdownDescription: "❗ Volume name.",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"size": schema.Int64Attribute{
+				MarkdownDescription: "Volume size in GB. If volume is attached, instance must be stopped to update this value. Volume can only grow, cannot be shrunk.",
```
pierre-emmanuelJ: The current situation is that the volume need to be detached from the instance to be resized, the instance can remain running

pkg/resources/block_storage/resource_volume.go
```diff
@@ -0,0 +1,588 @@
...
+	op, err := client.CreateBlockStorageVolume(ctx, request)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Unable to create Volume",
+			err.Error(),
+		)
+		return
+	}
+
+	_, err = client.Wait(ctx, op, exoscale.OperationStateSuccess)
```
pierre-emmanuelJ:
```suggestion
	op, err = client.Wait(ctx, op, exoscale.OperationStateSuccess)
```

Probably you can update the op variable here to make sure the ID reference is valid and set

pkg/resources/block_storage/resource_volume.go
```diff
@@ -0,0 +1,588 @@
...
+	if !plan.Labels.Equal(state.Labels) && !plan.Labels.IsUnknown() {
+		labels := exoscale.Labels{}
+
+		dg := plan.Labels.ElementsAs(ctx, &labels, false)
+		if dg.HasError() {
+			resp.Diagnostics.Append(dg...)
+			return
+		}
+
+		request := exoscale.UpdateBlockStorageVolumeLabelsRequest{
```
pierre-emmanuelJ: @arnaudgeiser is the `UpdateBlockStorageVolume` call already available?
Should we do like the CLI for now and wait for it?

pkg/resources/block_storage/resource_volume.go
```diff
@@ -0,0 +1,588 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"unable to parse volume ID",
+			err.Error(),
+		)
+		return
+	}
+
+	// Delete remote resource.
+	op, err := client.DeleteBlockStorageVolume(
```
pierre-emmanuelJ, line 509: Maybe we should check for attached instance first and then issue a detach call, because from an API perspective trying to delete a BS volume, will error out if an Instance is attached.

pierre-emmanuelJ (commented):
Good work @kobajagi :)
I'll try to do a second round later

pkg/resources/block_storage/resource_volume.go
kobajagi: I'll check, i am almost certain that I tried to delete the instance with attached volume and it worked.

pkg/resources/block_storage/resource_volume.go
kobajagi: Sure but description is still true. I can expand description to suggest both options:
- shutdown instance & resize volume;
- detach volume, resize it and re-attach it.

Not sure it is smart to do it automatically (detach -> resize -> attach) it will certainly break whatever workflow runs in the vm.

pkg/provider/provider.go
```diff
@@ -143,15 +147,35 @@ func (p *ExoscaleProvider) Configure(ctx context.Context, req provider.Configure
...
 	_ = clv2
 
+	// Exoscale v3 client
+	creds := credentials.NewStaticCredentials(
+		key,
+		secret,
+	)
+
+	opts := []exov3.ClientOpt{}
+	if ep := os.Getenv("EXOSCALE_API_ENDPOINT"); ep != "" {
```
sauterp, line 157: you're reusing `EXOSCALE_API_ENDPOINT` here, perhaps you can put it in a constant.

pkg/resources/instance/resource.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	egoscale "github.com/exoscale/egoscale/v2"
 	exoapi "github.com/exoscale/egoscale/v2/api"
+	egoscaleV3 "github.com/exoscale/egoscale/v3"
```
sauterp, line 17: could be called `exov3` like in other places

pkg/resources/instance/resource.go
```diff
@@ -349,6 +364,40 @@ func rCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag
 		}
 	}
 
+	// Attach blocks storage volumes if set
```
sauterp:
```suggestion
	// Attach block storage volumes if set
```

pkg/resources/instance/resource.go
```diff
@@ -349,6 +364,40 @@ func rCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag
...
+	if bsSet, ok := d.Get(AttrBlockStorageVolumeIDs).(*schema.Set); ok {
```
sauterp, line 368: I feel like this section could be in its own function

pkg/resources/instance/resource.go
```diff
@@ -484,6 +540,68 @@ func rUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag
...
+					return diag.Errorf("failed to detach block storage: %s", err)
+				}
+
+				_, err = clientV3.Wait(ctx, op, egoscaleV3.OperationStateSuccess)
+				if err != nil {
+					return diag.Errorf("failed to detach block storage: %s", err)
+				}
+			}
+		}
+	}
```
sauterp, line 607: Does this code also work if I add and remove volumes at the same time?

pkg/resources/instance/resource.go
```diff
@@ -484,6 +540,68 @@ func rUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag
...
+					request,
+				)
+				if err != nil {
+					return diag.Errorf("unable to parse attached instance ID: %s", err)
+				}
+
+				_, err = clientV3.Wait(ctx, op, egoscaleV3.OperationStateSuccess)
+				if err != nil {
+					return diag.Errorf("failed to attach block storage: %s", err)
+				}
```
sauterp, line 578: there might be some shared code between this section and the rCreate function that could be reused from a function.

sauterp (commented):
Looks very good, but it's a lot, I'll have another good look tomorrow. thanks~

pkg/resources/block_storage/resource_volume.go
pierre-emmanuelJ: Yes, we can expand the description, yes auto detach and attach could break some customer workload, I don't think it's convenient here you are right

pkg/resources/instance/resource.go
kobajagi: It makes sense, but i don't think it's worth the effort to refactor this resource as we are going to migrate the whole it to framework soon(ish).

pkg/resources/instance/resource.go
kobajagi: > It makes sense, but i don't think it's worth the effort to refactor this resource as we are going to migrate the whole it to framework soon(ish).

pkg/resources/block_storage/resource_volume.go
kobajagi: It's not in the prod openAPI spec.

pkg/resources/block_storage/resource_volume.go
```diff
@@ -0,0 +1,588 @@
...
+}
+
+// ResourceVolumeModel defines the resource data model.
+type ResourceVolumeModel struct {
+	ID             types.String `tfsdk:"id"`
+	Name           types.String `tfsdk:"name"`
+	Size           types.Int64  `tfsdk:"size"`
+	Blocksize      types.Int64  `tfsdk:"blocksize"`
+	CreatedAt      types.String `tfsdk:"created_at"`
+	Labels         types.Map    `tfsdk:"labels"`
```
sauterp, line 54: I think I found an issue related to this field, can it be fixed?

When I apply:
```hcl
resource "exoscale_block_storage_volume" "vol1" {
  name = "vol-1"
  size = 11
  zone = "ch-gva-2"
}
```

I get:
```shell
exoscale_block_storage_volume.vol1: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to exoscale_block_storage_volume.vol1, provider
│ "provider[\"registry.terraform.io/exoscale/exoscale\"]" produced an unexpected new value: .labels: was null, but now
│ cty.MapValEmpty(cty.String).
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

sauterp: it works if I specify `labels = {}`, I don't know if we want to "fix" this so it allows keeping it unset, or we could set the default value to `{}`.

pkg/resources/instance/resource.go
sauterp: nevermind, I tested it locally, works as intended.

sauterp (approved):
thanks, LGTM
leaving it up to you whether you want to change the default value of the `labels` field.

pkg/resources/block_storage/resource_volume.go
kobajagi: Fixed in https://github.com/exoscale/terraform-provider-exoscale/pull/341/commits/9c05a262e4d1176d922d961506d4bd5a3e428884

docs/resources/block_storage_volume.md
```diff
@@ -0,0 +1,57 @@
...
+Block Storage offers persistent externally attached volumes for your workloads.
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `name` (String) ❗ Volume name.
```
pierre-emmanuelJ, line 23: Just to notice here, there is an upcoming call `UpdateBlockStorageVolume ` to update the name/labels of a block storage volume, so in a near future this will be not immutable.
Shipping it immutable like that now, and update it after should not be a blocker or breaking change, no?

kobajagi: It will not break anything when we remove `RequreReplace`, so no problem here.

pkg/resources/block_storage/resource_volume.go
kobajagi: Fixed in https://github.com/exoscale/terraform-provider-exoscale/pull/341/commits/628800cbaaa8d94234f2577e09cc3fc315772940

pkg/resources/block_storage/resource_volume.go
kobajagi: I could remove option to update  labels in first release. Snapshot labels are also going to be immutable btw.

pkg/resources/block_storage/resource_volume.go
pierre-emmanuelJ: Let me check with the CLI

pkg/resources/block_storage/resource_volume.go
pierre-emmanuelJ: Snapshot will have a call also like for volume.
The issue if we add label update for now, if the API is updated the provider will break.
So let's not add the update on label for volume and snapshot for now.
Maybe we can do like in the CLI remove labels from this release (totally even at creation time)? Don't know what do you think?

pkg/resources/block_storage/resource_volume.go
kobajagi: Fixed in https://github.com/exoscale/terraform-provider-exoscale/pull/341/commits/0956eb80dd15d6502bd6cf0895dc1987276d2277

pkg/resources/block_storage/resource_volume.go
arnaudgeiser: The latest news (might diverge from what I said to some of you):

We can move forward with this and consider the `name` and `labels` fields immutable for the first version of the Terraform Provider.
Once the update commands are available on public-api (for volumes and snapshots), we make at least the `name` mutable. Giza will probably work at bringing those update commands to public-api.

kobajagi: Labels made immutable in https://github.com/exoscale/terraform-provider-exoscale/pull/341/commits/8c6c8c9521131e7a4bfdc352f029997d8d0433e0

pierre-emmanuelJ (approved):
🎱

sauterp (approved):

PR #367, opened 2024-06-07
Don't throw error if volume detach call fails due to volume not attached

kobajagi:
# Description

This PR fixes the race condition that may trigger error when both volume and volume attached instance are destroyed at the same time. As instance deletion automatically detaches volume, API throws error when detach is called as part of volume destroy.

Originally there was a check for volume state before detach is called, but that has proven unreliable. Here we are executing detach, but ignoring error if it says that volume is not attached.

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 BUG FIXES:
 
 - Fix panic when ListSKSClusterVersions returns error #357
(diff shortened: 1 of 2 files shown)


pierre-emmanuelJ (approved):
Sounds good to me for now

sauterp (approved):

Issue #395, opened 2024-11-20
[Bug]: Error with data exoscale_database_uri after upgrading to 0.61.1


maxlareo:
### Current Behavior

After upgrading from 0.61.0 to 0.61.1, the data exoscale_database_uri return the error: `Unable to fetch Database Service Postgres: RevealDBAASPostgresUserPassword: http response: Not Found: User avnadmin does not exist for service pg-name in ch-gva-2 zone`

I don't use `avnadmin` as admin user.

If I rollback to 0.61.0 I don't have the issue.

From comparing the version 0.61.0 and 0.61.1 I would say the issue comes from https://github.com/exoscale/terraform-provider-exoscale/blob/v0.61.1/pkg/resources/database/datasource_uri.go#L324  where `adminUsername` is set as `avnadmin` in https://github.com/exoscale/terraform-provider-exoscale/blob/master/pkg/resources/database/datasource_uri.go#L256.

### Expected Behavior

Don't have errors after upgrading exoscale terraform provider version.

### Steps To Reproduce

1. Pin exoscale provider version 0.61.0
1. Deploy a resource exoscale_database with an admin user different from default
1. Use a data exoscale_database_uri on the resource above
1. Upgrade to 0.61.1
1. Try to generate a plan

### Provider Version

0.61.1

### Terraform Version

OpenTofu v1.8.5

### Relevant log output

```shell
╷
│ Error: Client Error
│
│   with data.exoscale_database_uri.pg,
│   on main.tf line 38, in data "exoscale_database_uri" "pg":
│   38: data "exoscale_database_uri" "pg" {
│
│ Unable to fetch Database Service Postgres: RevealDBAASPostgresUserPassword: http response: Not Found: User avnadmin does not exist for service pg-name in ch-gva-2 zone
```

kobajagi:
Thank you @maxlareo for reporting the bug.
I can confirm it's a bug in the latest version, we are working on the fix.

maxlareo:
I tried again with terraform provider version 0.62.1 and it no longer hit the `Not Found: User avnadmin does not exist` error. So the issue is now fixed, thanks !

However before closing this issue, I now face a problem regarding authorization, looks like the provider now need an extra IAM policy rule to allow the new dedicated endpoint used to retrieve the dbaas credentials, thus it breaks the current IAM keys I'm using that does not allow the new resource in their policies.

Could you point me to the policy resource needed so I can create a new role policy that include it please ?

[...]

PR #396, opened 2024-11-20
dbaas: datasource username fix

simisoft-exo:
# Description
Fix allowing custom admin usernames, mentioned in https://github.com/exoscale/terraform-provider-exoscale/issues/395

## Checklist
(For exoscale contributors)

* [x] [Acceptance tests OK](https://github.com/exoscale/terraform-provider-exoscale/actions/runs/11972388480)

### DIFF
--- a/pkg/resources/database/datasource_uri.go
+++ b/pkg/resources/database/datasource_uri.go
@@ -553,4 +553,3 @@ func (d *DataSourceURI) Read(ctx context.Context, req datasource.ReadRequest, re
 
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
-
--- a/pkg/resources/database/datasource_uri.go
+++ b/pkg/resources/database/datasource_uri.go
@@ -255,17 +255,39 @@ func (d *DataSourceURI) Read(ctx context.Context, req datasource.ReadRequest, re
 
 	switch data.Type.ValueString() {
 	case "kafka":
-		res, err := client.GetDBAASServiceKafka(ctx, data.Name.ValueString())
+		res, err := waitForDBAASService(
+			ctx,
+			client.GetDBAASServiceKafka,
+			data.Name.ValueString(),
+			func(s *exoscale.DBAASServiceKafka) string { return string(s.State) },
+		)
 		if err != nil {
 			resp.Diagnostics.AddError(
 				"Client Error",
-				fmt.Sprintf("Unable to read Database Service kafka: %s", err),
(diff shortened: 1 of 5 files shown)


sauterp (approved):

pkg/resources/database/datasource_uri.go
```diff
@@ -283,9 +281,10 @@ func (d *DataSourceURI) Read(ctx context.Context, req datasource.ReadRequest, re
 			return
 		}
 
+		params = res.URIParams
 		data.Schema = types.StringValue("mysql")
 
-		creds, err := client.RevealDBAASMysqlUserPassword(ctx, data.Name.ValueString(), adminUsername)
+		creds, err := client.RevealDBAASMysqlUserPassword(ctx, data.Name.ValueString(), params["user"].(string))
```
kobajagi: Can you implement a check that `params` map has `user` key and that value is a string?
Although it's always expected to exist, in case of API hickup it will cause a panic in provider.

pkg/resources/database/datasource_uri.go
```diff
@@ -252,22 +252,47 @@ func (d *DataSourceURI) Read(ctx context.Context, req datasource.ReadRequest, re
...
+			if s, ok := i.(string); ok {
+				user = s
+			}
+		} else {
+			resp.Diagnostics.AddError(
+				"Client Error",
+				"Database Service Kafka parameter user is missing",
+			)
+			return
+		}
```
kobajagi: Way you implemented it only non-existing 'user' key is caught, but not non-string value (due to placement if else clause).
```suggestion
		if i, ok := params["user"]; ok {
			if s, ok := i.(string); ok {
				user = s
			}
		}
		if user == "" {
			resp.Diagnostics.AddError(
				"Client Error",
				"Database Service Kafka parameter user is missing",
			)
			return
		}
```

pkg/resources/database/datasource_uri.go
```diff
@@ -283,9 +308,21 @@ func (d *DataSourceURI) Read(ctx context.Context, req datasource.ReadRequest, re
...
+			if s, ok := i.(string); ok {
+				user = s
+			}
+		} else {
+			resp.Diagnostics.AddError(
+				"Client Error",
+				"Database Service MySQL parameter user is missing",
+			)
+			return
+		}
```
kobajagi: Same as https://github.com/exoscale/terraform-provider-exoscale/pull/396/files#r1852274114

kobajagi (approved):

PR #402, opened 2024-12-12
Docs fix

wdullaer:
# Description

* Uses `terraform` as language in code blocks. This results in better code highlighting in most renderers, including the one used on registry.terraform.io. It is also the language hint used by default by the docs framework.
* Fixes a number of broken links.

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK

### DIFF
--- a/pkg/resources/nlb_service/datasource_list.go
+++ b/pkg/resources/nlb_service/datasource_list.go
@@ -129,7 +129,7 @@ Corresponding resource: [exoscale_nlb](../resources/nlb.md).`,
 				Optional:            true,
 			},
 			NLBServiceListAttrNLBServiceList: schema.ListNestedAttribute{
-				MarkdownDescription: "The list of [exoscale_nlb_service](./nlb_service.md).",
+				MarkdownDescription: "The list of [exoscale_nlb_service](./nlb_service_list.md).",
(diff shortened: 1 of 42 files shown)


pkg/resources/nlb_service/datasource_list.go
```diff
@@ -129,7 +129,7 @@ Corresponding resource: [exoscale_nlb](../resources/nlb.md).`,
 				Optional:            true,
 			},
 			NLBServiceListAttrNLBServiceList: schema.ListNestedAttribute{
-				MarkdownDescription: "The list of [exoscale_nlb_service](./nlb_service.md).",
+				MarkdownDescription: "The list of [exoscale_nlb_service](./nlb_service_list.md).",
```
kobajagi, line 132: Not the best wording in description, but it should be `nlb_service.md` because it's a list of `nlb_service` elements.

wdullaer: I'm just doing a mechanical change to make all the links in the doc resolve with the current file names.

If we're not happy with the current file structure or descriptions, we can tackle that in a follow up PR.

kobajagi (approved):

PR #410, opened 2025-01-13
dbaas support for mysql,pg database

tgrondier:
# Description

New resource introduced:
- `exoscale_dbaas_pg_database`, `exoscale_dbaas_mysql_database`, to represent individual databases in database services.

To avoid confusion with the existing `exoscale_database` resource, it is being *deprecated*. A new resource is introduced, `exoscale_dbaas`. The only thing that changes is the name, under the hood those resources are exactly the same, except a deprecation message for the former.

(I couldn't find guidance on how to rename a resource type so I have no idea if it's possible to automate the move for users)

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

Acceptance tests updated


pkg/resources/database/resource_db_mysql.go
```diff
@@ -0,0 +1,286 @@
...
+		return
+	}
+	data.Timeouts = timeouts
+
+	data.Id = types.StringValue(databaseID)
+	data.DatabaseName = types.StringValue(databaseName)
+	data.Service = types.StringValue(serviceName)
+	data.Zone = types.StringValue(zone)
+
+	ReadResourceForImport(ctx, req, resp, &data, p.client)
```
kobajagi, line 94: I don't think this is needed, as stated in the [docs](https://developer.hashicorp.com/terraform/plugin/framework/resources/import) you only need to set required attributes for `Read` to work.

> Resources can implement the ImportState method, which must either specify enough Terraform state for the Read method to refresh [resource.Resource](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#Resource) or return an error.

pkg/resources/database/resource_db_mysql.go
```diff
@@ -0,0 +1,286 @@
...
+// UpdateResource updates the remote resource w/ the new model
+func (MysqlDatabaseResourceModel) UpdateResource(ctx context.Context, client *v3.Client, diagnostics *diag.Diagnostics) {
+	// Nothing to do as these resources are systematically recreated
+}
+
+func (data MysqlDatabaseResourceModel) WaitForService(ctx context.Context, client *v3.Client, diagnostics *diag.Diagnostics) {
+	_, err := waitForDBAASServiceReadyForFn(ctx, client.GetDBAASServiceMysql, data.Service.ValueString(), func(t *v3.DBAASServiceMysql) bool { return t.State == v3.EnumServiceStateRunning })
+	// DbaaS API is unstable when a service goes from rebuilding from running,
+	// this wait time helps avoid that
+	time.Sleep(time.Second * 10)
```
kobajagi: Put the sleep value as a var in utils.go so we can bump it at one place  if needed.

pkg/resources/database/resource_db_pg.go
```diff
@@ -0,0 +1,310 @@
...
+		return
+	}
+	data.Timeouts = timeouts
+
+	data.Id = types.StringValue(databaseID)
+	data.DatabaseName = types.StringValue(databaseName)
+	data.Service = types.StringValue(serviceName)
+	data.Zone = types.StringValue(zone)
+
+	ReadResourceForImport(ctx, req, resp, &data, p.client)
```
kobajagi, line 96: Not needed, see top comment.

pkg/resources/database/resource_db_pg.go
```diff
@@ -0,0 +1,310 @@
...
+func (data PGDatabaseResourceModel) ReadResource(ctx context.Context, client *v3.Client, diagnostics *diag.Diagnostics) {
+	svc, err := client.GetDBAASServicePG(ctx, data.Service.ValueString())
+	if err != nil {
+		diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read service pg, got error: %s", err))
+		return
+	}
+
+	for _, db := range svc.Databases {
+		if string(db) == data.DatabaseName.ValueString() {
+			return
```
kobajagi, line 200: Shouldn't you set `lc_collate` and `lc_ctype` here?

tgrondier: The information is not retrievable by the API so we don't know which values to update it to. I added a comment to make that clear.

pkg/resources/database/resource_db_mysql.go
tgrondier: There is a bit of confusion regarding the naming, ReadResourceForImport is a wrapper for `Read` but is named differently due to different signatures. I'll rename for clarity. (or I haven't understood correctly your comment)

kobajagi: I understand, what doc states is that `Import` should not call `Read` itself but only put enough data in the state that `Read` can be executed by terraform later. That usually translates to only setting attributes that are required for `Get*` API call to succeed.

tgrondier: Understood! I'll update and remove any unneeded attributes

kobajagi: After reading the rest of the code in this case it does not matter much as we only have required attributes so `Read` effectively just checks if resource exists. Still if `optional` attribute gets added latter we would have have to remove `Read` as we don't want to set optional value during `import`. Iit wouldn't hurt to update it now.

kobajagi (approved):

pierre-emmanuelJ (approved):

PR #412, opened 2025-03-03
sks cluster: enable_kube_proxy parameter

tgrondier:
# Description

https://openapi-v2.exoscale.com/operation/operation-create-sks-cluster#operation-create-sks-cluster-body-application-json-enable-kube-proxy

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

Local acceptance tests

### DIFF
--- a/exoscale/resource_exoscale_sks_cluster.go
+++ b/exoscale/resource_exoscale_sks_cluster.go
@@ -269,7 +269,7 @@ func resourceSKSClusterCreate(ctx context.Context, d *schema.ResourceData, meta 
 	}
 
 	createReq := v3.CreateSKSClusterRequest{
-		FeatureGates: make([]string, 0), // mandatory field but unimplemented in the provider yet
+		FeatureGates: make([]string, 0), // mandatory due to a bug https://github.com/exoscale/terraform-provider-exoscale/pull/412#discussion_r1977616102
 	}
 
 	var addOns []string
@@ -507,7 +507,7 @@ func resourceSKSClusterUpdate(ctx context.Context, d *schema.ResourceData, meta 
 
 	var updated bool
 	updateReq := v3.UpdateSKSClusterRequest{
-		FeatureGates: make([]string, 0), // mandatory field but unimplemented in the provider yet
+		FeatureGates: make([]string, 0), // mandatory field due to a bug https://github.com/exoscale/terraform-provider-exoscale/pull/412#discussion_r1977616102
 	}
 
 	if d.HasChange(resSKSClusterAttrAutoUpgrade) {
--- a/exoscale/resource_exoscale_sks_cluster.go
+++ b/exoscale/resource_exoscale_sks_cluster.go
@@ -506,9 +506,7 @@ func resourceSKSClusterUpdate(ctx context.Context, d *schema.ResourceData, meta 
 	}
 
 	var updated bool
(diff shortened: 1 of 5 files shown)


exoscale/resource_exoscale_sks_cluster_test.go
```diff
@@ -49,6 +49,8 @@ resource "exoscale_sks_cluster" "test" {
     test = "%s"
   }
 
+  enable_kube_proxy = true
```
nerzhul, line 52: can we have a test with it true and one with false ?

exoscale/resource_exoscale_sks_cluster.go
```diff
@@ -261,7 +268,9 @@ func resourceSKSClusterCreate(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
-	createReq := v3.CreateSKSClusterRequest{}
+	createReq := v3.CreateSKSClusterRequest{
+		FeatureGates: make([]string, 0), // mandatory field but unimplemented in the provider yet
```
sauterp: Consider this an informative comment, it doesn't matter for this PR, since this is a create request.

This field accepts "NULL" [link](https://openapi-v2.exoscale.com/operation/operation-create-sks-cluster#operation-create-sks-cluster-body-application-json-feature-gates). If it doesn't our doc is not up to date..

AFAIK most "empty fields"(slice of size zero, initialized but empty map etc.) are interpreted as in "clear this field", while "nil fields"(in this case for slices this is the zero or default value that is set when you don't set anything) are interpreted as in "don't touch this field". The latter is generally the better default IMHO.

So I would not set the `FeatureGates` here, thereby leaving it as `nil` and later once we start using the field, I would use `createReq.FeatureGates = append(createReq.FeatureGates, "whatever")`. `append` initializes the slice if it is nil.

sauterp (approved):

exoscale/resource_exoscale_sks_cluster.go
tgrondier: The field accepts null but the current version of egoscale doesn't; i should have been clearer about that. If we don't explicetly init this value in the request, we get the following response from the API:

At creation:
```
        Error: CreateSKSCluster: http response: Bad Request: Invalid value in feature-gates - should be a Collection
```

At updating:

```
        Error: UpdateSKSCluster: http response: Bad Request: bad request
```

It seems to be a bug in the API, I was going to investigate the issue further after this PR.

sauterp: thanks! yes I think that's a bug in the API too. It doesn't block this PR, but we should fix it before we implement update requests as we will want the "don't touch this field" behavior.

kobajagi (approved):

exoscale/resource_exoscale_sks_cluster_test.go
tgrondier: We can but it's not a meaningful test in that context: this attribute is only significant during cluster creation, and the result of it being set (or unset) is not checkable through the API.

So for now this will (temporarily) have to do.

exoscale/resource_exoscale_sks_cluster.go
tgrondier: Bug was resolved

nerzhul (approved):

sauterp (approved):

PR #413, opened 2025-03-07
Examples : add csi parameter in the example for SKS

SebastienPittet:
# Description
This PR follows a customer feedback. He used the example to create a SKS Cluster and then struggled with the PVC.
This, because the CSI addon is **NOT** deployed by default. 

Note : metrics_server and exoscale_ccm both default to true. 

As the auto_upgrade should not break, I suggest to set the boolean to true. 

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK


sauterp (approved):
Thanks! I agree this will serve most customers well.

nerzhul (approved):

PR #418, opened 2025-03-13
DBaaS: Add Valkey

simisoft-exo:


pkg/resources/database/resource_valkey.go
```diff
@@ -0,0 +1,294 @@
...
+		data.Name.ValueString(),
+		service,
+	)
+
+	if err != nil {
+		diagnostics.AddError("Client Error", fmt.Sprintf("Unable to create database service valkey, got error: %s", err))
+		return
+	}
+
+	r.readValkey(ctx, data, diagnostics)
```
kobajagi: This needs to be replaced with something similar to [this](https://github.com/exoscale/terraform-provider-exoscale/blob/master/pkg/resources/database/resource_pg.go#L204).

Instead of refreshing the whole resource, after create you should only refresh values that are:
- computed + not optional;
- computed + optional + not defined in the plan (`IsUnknown()==true`).

All attributes defined in plan must have keep values from the plan.

simisoft-exo: fixed in: https://github.com/exoscale/terraform-provider-exoscale/pull/418/commits/30b5a63a87273017a00ce05a08d310976798672f

kobajagi (approved):

kobajagi:
Looks good!
Just remember to update `CHANGELOG.md` before merging.

PR #427, opened 2025-04-18
Remove invalid ALL protocol in security group rule

aureliar8:
# Description
`ALL` is not a valid protocol to pass when creating a security group https://openapi-v2.exoscale.com/operation/operation-add-rule-to-security-group#operation-add-rule-to-security-group-body-application-json-protocol

Trying to use it is rejected by public API:
```tf
resource "exoscale_security_group_rule" "my_security_group_rule" {
  security_group_id = exoscale_security_group.tf-test.id
  type              = "INGRESS"
  protocol          = "ALL"
  cidr              = "0.0.0.0/0"
}
```

```sh
exoscale_security_group.tf-test: Creating...
exoscale_security_group.tf-test: Creation complete after 4s [id=03f8e0aa-786e-4fc9-a777-cad3079da861]
exoscale_security_group_rule.my_security_group_rule: Creating...
╷
│ Error: Post "https://api-ch-gva-2.exoscale.com/v2/security-group/03f8e0aa-786e-4fc9-a777-cad3079da861/rules": invalid request: Invalid value in protocol (description: `Network protocol`) - should be one of :ah, :esp, :gre, :icmp, :icmpv6, :ipip, :tcp, :udp
│ 
│ 
│   with exoscale_security_group_rule.my_security_group_rule,
│   on main.tf line 19, in resource "exoscale_security_group_rule" "my_security_group_rule":
│   19: resource "exoscale_security_group_rule" "my_security_group_rule" {
│ 
╵
```

According to git blame this `"ALL"` protocol in the terraform provider predates the switch to the V2 API. Maybe it used to be possible to set this protocol in the V1 api, but it's not the case anymore
## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

### DIFF
--- a/docs/resources/security_group_rule.md
+++ b/docs/resources/security_group_rule.md
@@ -44,7 +44,7 @@ directory for complete configuration examples.
 - `end_port` (Number) ❗ A `TCP`/`UDP` port range to match.
 - `icmp_code` (Number) ❗ An ICMP/ICMPv6 [type/code](https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol#Control_messages) to match.
 - `icmp_type` (Number) ❗ An ICMP/ICMPv6 [type/code](https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol#Control_messages) to match.
-- `protocol` (String) ❗ The network protocol to match (`TCP`, `UDP`, `ICMP`, `ICMPv6`, `AH`, `ESP`, `GRE`, `IPIP` or `ALL`)
+- `protocol` (String) ❗ The network protocol to match (`TCP`, `UDP`, `ICMP`, `ICMPv6`, `AH`, `ESP`, `GRE` or  `IPIP`)
 - `public_security_group` (String) ❗ An (`INGRESS`) source / (`EGRESS`) destination public security group name to match (conflicts with `cidr`/`user_security_group`/`user_security_group_id`).
 - `security_group` (String, Deprecated) ❗ The parent security group name. Please use the `security_group_id` argument along the [exoscale_security_group](../data-sources/security_group.md) data source instead.
 - `security_group_id` (String) ❗ The parent [exoscale_security_group](./security_group.md) ID.
--- a/exoscale/resource_exoscale_security_group_rule.go
+++ b/exoscale/resource_exoscale_security_group_rule.go
@@ -36,6 +36,10 @@ const (
 
 var securityGroupRuleProtocols = []string{
 	"AH",
+	// Even though public API don't support creating ALL rules. Some customer might have terraform
+	// code that references such rule that was created via obwald. So we need to keep support for it
+	// to not break customers.
+	"ALL",
 	"ESP",
 	"GRE",
 	"ICMP",
@@ -119,7 +123,7 @@ func resourceSecurityGroupRule() *schema.Resource {
 				ValidateFunc: validation.StringInSlice(securityGroupRuleProtocols, true),
 				// Ignore case differences
 				DiffSuppressFunc: suppressCaseDiff,
-				Description:      "The network protocol to match (`TCP`, `UDP`, `ICMP`, `ICMPv6`, `AH`, `ESP`, `GRE` or `IPIP`)",
+				Description:      "The network protocol to match (`TCP`, `UDP`, `ICMP`, `ICMPv6`, `AH`, `ESP`, `GRE`, `IPIP` or `ALL`)",
 			},
 			resSecurityGroupRuleAttrPublicSecurityGroup: {
 				Type:        schema.TypeString,
--- a/exoscale/resource_exoscale_security_group_rule.go
+++ b/exoscale/resource_exoscale_security_group_rule.go
@@ -36,7 +36,7 @@ const (
 
 var securityGroupRuleProtocols = []string{
 	"AH",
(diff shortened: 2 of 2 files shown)


kobajagi (approved):

tgrondier (approved):

ste-m:
uhm... @aureliar8 i need to investigate this a bit. I'm under the impressions we stumbled upon it when migrating from Obwald to Pubic API, and as you are doing right now, at the time we went initially for the radical approach, just to be recalled to reality by customers tickets.

ste-m:
So here is the thing: Public API is surely declaring an OpenAPI spec without `all` since at least 11 months:
https://github.com/exoscale/public-api/commit/6f177769d594c17c5339bd67a17623987c313376
That is roughly when all things got moved around in PublicAPI.

And from a bit of investigation it appears that `:all` was absent from first implementation:
https://github.com/exoscale/schemas/commit/2d7d93b9ed38a679d4f4d3b65e2bacffaa0b4319

We have a trace of the `:all` keyword being considered missing here:
https://app.shortcut.com/exoscale/story/16224/public-api-security-groups

And with the power of IRC logs:

```
[2020-07-08T09:21:54+0200] <ste> hello, i'm searching for a sec groups expert :)
[2020-07-08T09:21:54+0200] <ste> i have a nice SG rule that has been created with an "all" keyword for the protocol
[2020-07-08T09:21:54+0200] <ste> https://skathi.sos-ch-dk-2.exo.io/2020-07-07_17.10.39.png
[2020-07-08T09:21:54+0200] <ste> seems that this is an accepted value for the CS API
[2020-07-08T09:21:54+0200] <ste> who knows how that would behave for eg ICMP?
[2020-07-08T09:21:54+0200] <ste> is :all an interesting value to be passed for protocol, and should we consider it in our tooling and in public-api?
[2020-07-08T09:32:55+0200] <mathieu> ste: I was not aware of this "all" parameter
[2020-07-08T09:33:05+0200] <ste> mathieu me neither ‾\_(ツ)_/‾ 
[2020-07-08T09:33:13+0200] <ste> this is why i'm investigating
[2020-07-08T09:33:39+0200] <ste> so we can do an educated choice in implementing sec grops in the public api
[2020-07-08T09:33:42+0200] <ste> *groups

```

You can see how i had an SG rule `:all` in the referenced image:
https://skathi.sos-ch-dk-2.exo.io/2020-07-07_17.10.39.png

Always with the power if IRC logs i can find this guy here, which gives us more clues:
```
[2021-03-11T12:09:08+0100] <sheldon> schemas mcorbin pushed 1 new commit to mcorbin/ch26146/public-api-pre-prod-list-security-groups: Security groups: :all protocol https://git.io/JqnBG
```

Which brings us here:
https://github.com/exoscale/schemas/commit/7f9f4a2bb87f

I read this last one as:

- impede user to create `:all` rules
- allow for existing `:all` rules to be functional

And in fact, the current picture is:

https://bi.internal.exoscale.ch/question/4117-security-group-rule-filtered-by-protocol-is-all

I conclude that the TF implementation of `:all` **might** be required for existing `:all` rules, and the removal of the `:all` implementation **might** impact existing `:all` rules if managed via TF.

Using the "might" verb as i have no real factual clue.

ste-m:
Side note: we might want to reach out and force those customers to convert their rules, or convert them ourselves forcibly in order to fully deprecate `:all`.

arnaudgeiser (changes requested):
With the information I have at the moment, I don't think it should be merged.

aureliar8:
Some customer that created `ALL` rules via obwald need to keep support to refer to `ALL` rules in their terraform setup, even if new ones can't be created. 

I added back support for it inside the provider. This PR is now only a documentation update to stop advertising that new `ALL` rules can be created

arnaudgeiser (approved):
Thanks aurelien!

arnaudgeiser:
> Some customer that created ALL rules via obwald need to keep support to refer to ALL rules in their terraform setup, even if new ones can't be created.

I'm wondering.. didn't we introduce new Exoscale datasources when moving from Obwald to Public-API and actually no "ALL" security-group rules could have been created with this one?
If the answer is yes, it would probably mean the original proposition of aurelien can be considered.

cc @kobajagi

kobajagi:
@arnaudgeiser don't think that is true. We deprecated `exoscale_security_group_rules` resource in favor of `exoscale_security_group_rule` (which requires one resource definition per SG rule. However I can see in the git history that both used obwald originally (see [this commit](https://github.com/exoscale/terraform-provider-exoscale/commit/0130e246b7421b136c2726ed244e67c32866bf13)). So it is plausable we have a user that created SG rule with Obwald and `ALL` protocol and would get error if we remove `ALL` from the list of allowed protocols.

arnaudgeiser:
> @arnaudgeiser don't think that is true. We deprecated exoscale_security_group_rules resource in favor of exoscale_security_group_rule (which requires one resource definition per SG rule. However I can see in the git history that both used obwald originally (see [this commit](https://github.com/exoscale/terraform-provider-exoscale/commit/0130e246b7421b136c2726ed244e67c32866bf13)). So it is plausable we have a user that created SG rule with Obwald and ALL protocol and would get error if we remove ALL from the list of allowed protocols.

Snif but thanks!

pierre-emmanuelJ (approved):
good catch

PR #430, opened 2025-05-06
Acceptance tests with action args

simisoft-exo:
# Description

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,12 @@ name: run-acceptance-tests
 
 on:
   workflow_dispatch:
+    inputs:
+      go_test_extra_args:
+        description: 'Extra arguments to pass to Go tests (e.g. -run "^Test(?!.*(Database|SOS)).*")'
+        required: false
+        type: string
+        default: ''
   push:
     branches:
     - '**'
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -27,8 +33,16 @@ jobs:
 
       - uses: ./go.mk/.github/actions/pre-check
 
-      - name: Run tests
+      - name: Run tests with default arguments
+        if: github.event.inputs.go_test_extra_args == ''
```
sauterp: Nice idea! But do we really need this `if`? Isn't it enough to leave GO_TEST_EXTRA_ARGS empty and it will run all the tests?

simisoft-exo: There is a default in the makefile, wouldn't leaving it empty set it to an empty string and would be an override. Not sure what's the workflow behavior, so I decided to be explicit.

pierre-emmanuelJ (approved):

.github/workflows/tests.yml
kobajagi: I think empty string  would not override Makefile default as empty env var is usually handled same as unset. It's worth testing to be certain.

.github/workflows/tests.yml
```diff
@@ -24,11 +30,19 @@ jobs:
...
+          EXOSCALE_API_KEY: ${{ secrets.EXOSCALE_API_KEY }}
+          EXOSCALE_API_SECRET: ${{ secrets.EXOSCALE_API_SECRET }}
+      - name: Run tests with custom arguments
+        if: github.event.inputs.test_run_pattern != ''
+        run: |
+          make test-verbose test-acc EXTRA_ARGS="-parallel=3 -count=1 -failfast -run=${{ github.event.inputs.test_run_pattern }}"
         env:
           EXOSCALE_API_KEY: ${{ secrets.EXOSCALE_API_KEY }}
           EXOSCALE_API_SECRET: ${{ secrets.EXOSCALE_API_SECRET }}
+          GO_TEST_EXTRA_ARGS: ${{ github.event.inputs.test_run_pattern }}
```
sauterp: hmm, now you are setting the test_run_pattern in two places for this step, once as an env var and once in the EXTRA_ARGS on line 44. What's the reason for this?

Also isn't it possible to combine this step and the previous one? AFAICT we just need to remove line 35(the if) and move line 48 after line 40. Then I think we're good :)

simisoft-exo: This is the way I got make precedence to play ball is to explicitly pass the var. The last one is ignored by make, so I removed it.

sauterp (approved):
thanks!

PR #458, opened 2025-09-01
Add support for Kubernetes Audit

remyrd:
# Description

Add support for Kubernetes Audit - Create / Update / Read

Taken by @nerzhul 

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

```
❯ TF_ACC=1 go test ./... -run 'TestAccResourceSKSCluster$'            
?   	github.com/exoscale/terraform-provider-exoscale	[no test files]
ok  	github.com/exoscale/terraform-provider-exoscale/exoscale	291.774s
?   	github.com/exoscale/terraform-provider-exoscale/pkg/config	[no test files]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/filter	(cached) [no tests to run]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/general	[no test files]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/list	(cached) [no tests to run]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/provider	[no test files]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/provider/config	[no test files]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/anti_affinity_group	(cached) [no tests to run]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/block_storage	(cached) [no tests to run]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/database	(cached) [no tests to run]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/iam	(cached) [no tests to run]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance	(cached) [no tests to run]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance_pool	(cached) [no tests to run]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/nlb_service	(cached) [no tests to run]
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/sos_bucket_policy	(cached) [no tests to run]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/resources/zones	[no test files]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/sos	[no test files]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/testutils	[no test files]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/utils	[no test files]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/validators	[no test files]
?   	github.com/exoscale/terraform-provider-exoscale/pkg/version	[no test files]
?   	github.com/exoscale/terraform-provider-exoscale/version	[no test files]

```


exoscale/resource_exoscale_sks_cluster.go
```diff
@@ -76,6 +81,37 @@ func resourceSKSCluster() *schema.Resource {
...
+			Optional: true,
+			Computed: true,
+			Description: "Parameters for Kubernetes Audit configuration (may only be enabled at creation time)",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema {
+				    resSKSClusterAttrAuditEnabled: {
+						Type: schema.TypeBool,
+						Required: false,
+						Description: "Whether to run the APIServer with the configured Kubernetes Audit",
+					},
```
PhilippeChepy: Once sc-135074 is implemented, this boolean may be pointless (customers provides their config or not, if there is any, it's enabled, otherwise it's not).

remyrd: Unfortunately this clashes with the portal which can send null values (it didn't cross my mind when we discussed it). If it doesn't hurt i would keep it for now. We can add some ad-hoc logic to provide this behavior

remyrd: eg. no config = `"audit": {"enabled": false}`

remyrd: > Once sc-135074 is implemented, this boolean may be pointless (customers provides their config or not, if there is any, it's enabled, otherwise it's not).

i was thinking about this and that's not how the api works more generally. abscence of a field during update means we don't touch it. we don't pass the whole cluster to update as a payload, only what needs change.

Since we offer to enable/disable the feature, we need to make the difference between an update that doesn't touch kube-audit, and one that disables it. If we want to do this via tf, we can detect change and send `{"audit": {"enabled": false}}` when we detect the customer deleted the field. wdyt?

exoscale/resource_exoscale_sks_cluster.go
```diff
@@ -76,6 +81,37 @@ func resourceSKSCluster() *schema.Resource {
 			Computed:    true,
 			Description: "The CA certificate (in PEM format) for TLS communications between the control plane and the aggregation layer (e.g. `metrics-server`).",
 		},
+		"audit": {
+			Type:        schema.TypeList,
+			MaxItems:    1,
+			Optional:    true,
+			Computed:    true,
+			Description: "Parameters for Kubernetes Audit configuration (may only be enabled at creation time)",
```
remyrd, line 88:
```suggestion
			Description: "Parameters for Kubernetes Audit configuration",
```

remyrd (commented):
I'm not an expert here but so far looks good to me. The only detail worth mentioning is when updating the cluster, the abscence of an audit config in the payload :warning: **does not** :warning: disable it. The user has to explicitly set `enabled: false`

exoscale/resource_exoscale_sks_cluster.go
```diff
@@ -76,6 +81,37 @@ func resourceSKSCluster() *schema.Resource {
 			Computed:    true,
 			Description: "The CA certificate (in PEM format) for TLS communications between the control plane and the aggregation layer (e.g. `metrics-server`).",
 		},
+		"audit": {
+			Type:        schema.TypeList,
+			MaxItems:    1,
+			Optional:    true,
+			Computed:    true,
```
kobajagi: You did not implement it as computed because you are not reading the value from the server (`resourceSKSClusterRead`). You can remove this line.

exoscale/resource_exoscale_sks_cluster.go
```diff
@@ -76,6 +81,37 @@ func resourceSKSCluster() *schema.Resource {
...
+			Description: "Parameters for Kubernetes Audit configuration",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					resSKSClusterAttrAuditEnabled: {
+						Type:        schema.TypeBool,
+						Required:    false,
```
kobajagi: `false` is default, you can remove the line.

kobajagi: Actually replace it with `Optional: true` because that is how it is implemented.

exoscale/resource_exoscale_sks_cluster.go
```diff
@@ -76,6 +81,37 @@ func resourceSKSCluster() *schema.Resource {
...
+						Description: "Whether to run the APIServer with the configured Kubernetes Audit",
+					},
+					resSKSClusterAttrAuditEndpoint: {
+						Type:        schema.TypeString,
+						Required:    true,
+						Description: "The Endpoint URL for the Webserver responsible of processing Audit events",
+					},
+					resSKSClusterAttrAuditInitBackoff: {
+						Type:        schema.TypeString,
+						Required:    false,
```
kobajagi: Same as https://github.com/exoscale/terraform-provider-exoscale/pull/458/files#r2381767084

exoscale/resource_exoscale_sks_cluster.go
```diff
@@ -76,6 +81,37 @@ func resourceSKSCluster() *schema.Resource {
...
+						Description: "The Initial Backoff to wait before sending data to the remote server (default '10s')",
+					},
+					resSKSClusterAttrAuditBearerToken: {
+						Type:        schema.TypeString,
+						Required:    false,
```
kobajagi: Same as https://github.com/exoscale/terraform-provider-exoscale/pull/458/files#r2381767084

kobajagi:
> I'm not an expert here but so far looks good to me. The only detail worth mentioning is when updating the cluster, the abscence of an audit config in the payload ⚠️ **does not** ⚠️ disable it. The user has to explicitly set `enabled: false`

That is expected behavior of an optional attribute in terraform. Adding a note never hurts.

nerzhul:
there is some acceptance tests to fix, some are out of this PR scope, anyway

kobajagi (approved):

PhilippeChepy (approved):

PR #461, opened 2025-10-06
Make instance.ssh_key(s) force resource replace

kobajagi:
# Description

As we only support setting SSH key(s) on instance create, this PR marks the corresponding attribute with `ForceNew` to force resource replacement.

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK


pierre-emmanuelJ (approved):
🙏

sauterp (approved):

PR #470, opened 2025-11-17
CI: make check-documentation-clean.sh fail when `go generate` fails

sauterp:
# Description
Previously the CI step that checks that the documentation was up to date didn't fail when `go generate` failed. It only failed when there was a git diff afterwards. We fix that.

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK

## Testing
```shell
bash-5.3$ ./scripts/check-documentation-clean.sh 
main.go:18: running "terraform": exec: "terraform": executable file not found in $PATH
documentation is up to date
bash-5.3$ echo $?
0
```
After adding `set -e`:
```shell
bash-5.3$ ./scripts/check-documentation-clean.sh 
main.go:18: running "terraform": exec: "terraform": executable file not found in $PATH
bash-5.3$ echo $?
1
```

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,5 +17,6 @@ jobs:
       - uses: actions/checkout@v3
       - run: make go.mk
       - uses: ./go.mk/.github/actions/setup
+      - uses: hashicorp/setup-terraform@v3
       - name: Check documentation is up to date
         run: './scripts/check-documentation-clean.sh'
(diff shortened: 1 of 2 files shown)


kobajagi (approved):

aureliar8 (approved):
I always try to use `set -euo pipefail` in my bash scripts 
http://redsymbol.net/articles/unofficial-bash-strict-mode/

nerzhul (approved):

pierre-emmanuelJ (approved):

PR #475, opened 2025-12-01
docs(instance): fix docs of state attribute

gzm0:
# Description

It is optional and does not have a default.

## Checklist

* [ ] Acceptance tests OK

## Testing

None


kobajagi (approved):

kobajagi:
On a second thought, default state (`running`) is implied on API side: if `status` is not specified in plan then still instance will start.
So need to reword this, not remove. I'll create a new PR for that.
Thank you for bringing it to our attention.

PR #489, opened 2026-01-26
feat: add private attribute to instance_pool resource

mdnix:
# Description

We bootstrap our infrastructure via this provider and need private instance pools. The option was available in the web UI and [API](https://community.exoscale.com/reference/api/compute/instance-pool/#request-body) but not exposed in this provider.

I added the `private` attribute to `exoscale_instance_pool` resource to support creating instance pools without public IP addresses. This matches the existing behavior of `exoscale_compute_instance`.

```
resource "exoscale_instance_pool" "workers" {
  # ...
  private = true
}
```

## Checklist

* [ ] Acceptance tests OK

## Testing

Tested locally with OpenTofu using dev_overrides in ~/.tofurc - successfully created a private instance pool


mdnix:
any input here would be great

kobajagi:
Thanks for contribution. UX vise i am leaning toward adding a new attribute `public-ip` instead (done already in [cli](https://github.com/exoscale/cli/blob/master/CHANGELOG.md#breaking-changes-1)). As is `private` is in conflict with `ipv6_address` and should be marked as such.

Anyway this feature is on our roadmap, once we internally decide on the approach we may base on your PR so I'll keep it open for a bit longer. Thanks again.

mdnix:
> Thanks for contribution. UX vise i am leaning toward adding a new attribute `public-ip` instead (done already in [cli](https://github.com/exoscale/cli/blob/master/CHANGELOG.md#breaking-changes-1)). As is `private` is in conflict with `ipv6_address` and should be marked as such.
> 
> Anyway this feature is on our roadmap, once we internally decide on the approach we may base on your PR so I'll keep it open for a bit longer. Thanks again.

Thanks for the update. Glad this is on the roadmap. Happy to help out here

PR #492, opened 2026-02-02
instance_pool: Add support for private instance pools

francoismeulenberg:
# Description

**instance_pool:** Add support for private instance pools (no public IP addresses)

### Implementation
- Added `AttrPrivate` schema attribute with `ForceNew: true` (private pools cannot be converted to public or vice versa)
- Implemented mapping between `private`/`ipv6` attributes and the API's `PublicIPAssignment` field:
  - `private = true` → `PublicIPAssignment = "none"` (no public IP)
  - `ipv6 = true` → `PublicIPAssignment = "dual"` (IPv4 + IPv6)
  - Default → `PublicIPAssignment = "inet4"` (IPv4 only)
- Added mutual exclusivity constraint: `private` and `ipv6` cannot both be enabled (ConflictsWith validation)
- Updated create, read, and update functions to handle the new PublicIPAssignment logic
- Generated documentation reflecting the new attribute

Comments for This PR

**On `rCreate()` PublicIPAssignment logic:**
> Mapping between Terraform attributes and API v3 PublicIPAssignment field.
> This matches the pattern used in the exoscale_instance resource.

**On `AttrPrivate` schema:**
> ForceNew: true because API doesn't allow changing public IP assignment after pool creation.
> ConflictsWith ipv6 since they map to different PublicIPAssignment values.

**On `rUpdate()` IPv6 handling:**
> Only update PublicIPAssignment if not private, since private pools have
> PublicIPAssignment="none" which cannot be changed (ForceNew).

**On `testResourcePrivate()` cloud-init:**
> Cloud-init configures netplan for eth1 (private network interface) with DHCP.
> Required for private instances to get an IP and successfully boot.

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

- Added `testResourcePrivate()` acceptance test that:
  - Creates a private instance pool with a private network attachment
  - Includes cloud-init configuration to set up DHCP on the private network interface (eth1)
  - Verifies `PublicIPAssignment = "none"` is set correctly in the API
  - Validates instances boot successfully and pool reaches healthy state
- All tests pass successfully:
  - `TestInstancePool/ResourcePrivate`
  - Original `TestInstancePool/Resource` remains unchanged and passing
 
```shell
➜  terraform-provider-exoscale git:(make-instance-pools-private) ✗ TF_ACC=1 go test -v -timeout=90m -tags=testacc ./pkg/resources/instance_pool/                                      
=== RUN   TestInstancePool
=== RUN   TestInstancePool/DataSource
=== RUN   TestInstancePool/DataSourceList
=== RUN   TestInstancePool/Resource
=== RUN   TestInstancePool/ResourcePrivate
--- PASS: TestInstancePool (242.19s)
    --- PASS: TestInstancePool/DataSource (59.27s)
    --- PASS: TestInstancePool/DataSourceList (38.75s)
    --- PASS: TestInstancePool/Resource (96.63s)
    --- PASS: TestInstancePool/ResourcePrivate (47.54s)
PASS
ok      github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance_pool     242.216s
```


francoismeulenberg:
Background information:
For my current Exoscale project I need to be able to create private instance pools.
I noticed that this option was not yet available in the Exoscale provider for instance pools.

So I have added it.

pkg/resources/instance_pool/resource.go
```diff
@@ -350,8 +359,31 @@ func rCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag
...
+	} else if enableIPv6, ok := d.GetOk(AttrIPv6); ok {
+		ipv6EnabledBool := enableIPv6.(bool)
+		if ipv6EnabledBool {
+			createPoolRequest.PublicIPAssignment = v3.CreateInstancePoolRequestPublicIPAssignmentDual
+		} else {
+			createPoolRequest.PublicIPAssignment = v3.CreateInstancePoolRequestPublicIPAssignmentInet4
+		}
+	} else {
+		createPoolRequest.PublicIPAssignment = v3.CreateInstancePoolRequestPublicIPAssignmentInet4
+	}
```
kobajagi, line 386: Can you simplify this block with `switch` please?

kobajagi: hint: you can default to ipv4.

pkg/resources/instance_pool/resource.go
```diff
@@ -707,8 +747,29 @@ func rApply(ctx context.Context, client *v3.Client, d *schema.ResourceData, pool
...
+			return diag.FromErr(err)
+		}
+	} else {
+		// inet4 or empty
+		if err := d.Set(AttrPrivate, false); err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set(AttrIPv6, false); err != nil {
+			return diag.FromErr(err)
+		}
```
quentinalbertone, line 772: **suggestion:** I would prefer if you:
* use a switch here
* use only once the `d.Set(AttrPrivate,...)` and `d.Set(AttrIPv6, ...)`, it will keep the code easy to read (you can use variables in your switch case and set variable in the state ouside of the switch)

pkg/resources/instance_pool/resource_test.go
```diff
@@ -100,7 +101,7 @@ locals {
 
 data "exoscale_template" "debian" {
   zone = local.zone
-  name = "Linux Debian 12 (Bookworm) 64-bit"
+  name = "Linux Ubuntu 24.04 LTS 64-bit"
```
kobajagi, line 104: Nitpick but resource name (2 lines above) is `debian`.

pkg/resources/instance_pool/resource_test.go
```diff
@@ -153,6 +154,61 @@ resource "exoscale_instance_pool" "test" {
...
+            version: 2
+            ethernets:
+              eth1:
+                dhcp4: true
+
+    runcmd:
+      - [ netplan, apply ]
+  EOT
+
+  depends_on = [exoscale_private_network.test_private]
```
kobajagi, line 198: Should not be needed

kobajagi: terraform will make it depend due to reference in `network_ids`

Issue #511, opened 2026-04-01
[Feature Request]: Support DBaaS read replicas


owngr:
### What feature would you like to have in the provider?

Hello,

We would like to be able to create read replicas for Postgres with Terraform/OpenTofu, as it's possible with exoscale cli:
https://community.exoscale.com/product/dbaas/service-specific/postgresql/how-to/read-replicas/

This would help us automate the deployment from end to end with OpenTofu.

kobajagi:
Thanks @owngr , it is already on our roadmap, i will update the ticket when we start working on it.

yenda:
I needed it as well so I implemented it in my fork if you are interested I opened a PR 

[PR](https://github.com/exoscale/terraform-provider-exoscale/pull/514)

I took a more terraform idiomatic approach than aiven provider, since it was the way the exoscale provider is doing things.
It is AI generated/human directed. I ran all the tests against a test account.

loics2:
@kobajagi do you have an estimate of the timeline for this feature? It would be really useful to us

PR #514, opened 2026-04-10
feat(dbaas): support integrations on pg and mysql services

yenda:
## Summary

Adds an `integrations` SetNestedAttribute to the `pg` and `mysql` DBaaS service resources, allowing operators to declare read replicas at creation time.

Implements #511.

```hcl
resource "exoscale_dbaas" "replica" {
  name = "my-replica"
  type = "pg"
  plan = "startup-4"
  zone = "ch-gva-2"
  pg = {
    version = "15"
    integrations = [{
      type           = "read_replica"
      source_service = exoscale_dbaas.primary.name
    }]
  }
}
```

## Design

- **Schema**: `SetNestedAttribute` with two Required string children (`type`, `source_service`). Optional+Computed with custom `setUseStateForUnknown` and `setRequiresReplace` plan modifiers (the upstream `setplanmodifier` package is not vendored in this repo). `integrationsSelfSource` validator rejects configurations where `source_service` points at the declaring resource.
- **Create path**: Three-layer unknown-value defense (plan-time validators skip unknowns gracefully; create-time rejects whole-attribute-unknown by comparing plan vs config; per-element unknown/null values and semantic checks re-run after interpolations resolve). Orphan cleanup via defer+flag deletes the service if post-create steps fail.
- **Read path**: Only surfaces integrations where `integration.Dest` matches the current service, so a primary doesn't pick up replicas.
- **Update path**: Defensive no-op sync (RequiresReplace means Terraform never calls Update for integration changes).

## Additional fixes

- Fix pre-existing bug in `updateMysql` where `backup_schedule` parsing lacked `IsUnknown()` guards (unlike the pg equivalent), causing failures when Optional+Computed attributes are omitted from config during Update.

## Test coverage

| Test | Type | What it covers |
|---|---|---|
| `testResourcePgIntegrations` | Acceptance (live API) | Create replica, import, out-of-band delete detection, swap source (Replace), omit integrations (Update, not Replace) |
| `testResourceMysqlIntegrations` | Acceptance (live API) | Same lifecycle for mysql |
| `testResourcePgIntegrationsValidators` | Plan-only (zero API cost) | Empty set, bad type, self-source, multi-element |
| `testResourcePgIntegrationsUnknownSource` | Plan-only (zero API cost) | Unknown nested values don't break plan |
| `testResourcePgStateUpgrade` | Acceptance (live API) | v0.68.0 → current provider upgrade with no drift |

## Files changed

- `pkg/resources/database/planmodifiers.go` — new file: custom Set plan modifiers + self-source validator
- `pkg/resources/database/resource_service.go` — pass config to createPg/createMysql for unknown detection
- `pkg/resources/database/resource_service_pg.go` — schema, create/read/update integration handling
- `pkg/resources/database/resource_service_mysql.go` — same for mysql + backup_schedule fix
- `pkg/resources/database/resource_service_pg_test.go` — 5 new test functions
- `pkg/resources/database/resource_service_mysql_test.go` — mysql integration test
- `pkg/resources/database/main_test.go` — register new subtests
- `pkg/resources/database/testdata/resource_pg.tmpl` — template support for integrations + depends_on
- `pkg/resources/database/testdata/resource_mysql.tmpl` — same
- `docs/resources/database.md`, `docs/resources/dbaas.md` — regenerated
- `CHANGELOG.md`, `README.md` — feature entry + OpenTofu instructions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ BREAKING CHANGES:
 FEATURES:
 
 - feat: call crossplane generation after a new release #509
+- feat(dbaas): support `integrations` attribute on pg and mysql services to declare read replicas at creation time #511.
+  The attribute is Optional, so existing services are unaffected by the upgrade (they refresh with `integrations = null`).
+  Only integrations where **this** resource is the destination are exposed (e.g. declare the block on the replica, not the primary).
+  Changes to the set force resource replacement — the database service is destroyed and recreated, including all data on it.
+  Removing an integration out-of-band (e.g. via the Exoscale dashboard) will also trigger a forced replace on the next plan.
+  Only `read_replica` integrations are supported inline today; other integration types must still be managed via the standalone API endpoint.
 
 ## 0.68.0
 
--- a/README.md
+++ b/README.md
@@ -84,6 +84,24 @@ website][tf-doc].
 make GO_TEST_EXTRA_ARGS="-v -run ^TestAccResourceCompute$" test-acc
 ```
 
+#### Running acceptance tests under OpenTofu
+
+The provider is compatible with both Terraform and [OpenTofu](https://opentofu.org/).
+To run the acceptance tests against OpenTofu instead of Terraform, point
+`terraform-plugin-testing` at the `tofu` binary and override the default
+provider source host — the test harness defaults to
+`registry.terraform.io/-/exoscale` which OpenTofu rejects with a parse
+error on the legacy `-` namespace:
+
+```sh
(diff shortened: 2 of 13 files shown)


simisoft-exo:
Thank you for your suggestion.
This design is not aligned with our internal guidelines.
Aiven treats integrations and read-replicas the same but in an inconsistent way across endpoints.
Conceptually we treat read-replicas as different from integrations and would have to be more in line to how things look like in our Portal.
We added it to our roadmap to implement it as a subresource.

PR #522, opened 2026-04-22
Support for SKS default security group

PhilippeChepy:
# Description

- SKS: Add support for automatic security group creation
- SKS: Export the automatic security group id
- Update related documentation

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

CI

### DIFF
--- a/exoscale/resource_exoscale_sks_cluster_test.go
+++ b/exoscale/resource_exoscale_sks_cluster_test.go
@@ -72,6 +72,7 @@ resource "exoscale_sks_cluster" "test" {
   feature_gates = ["%s"]
 
   enable_kube_proxy = true
+  create_default_security_group = true
 
   oidc {
     client_id  = "%s"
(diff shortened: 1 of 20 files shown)


exoscale/resource_exoscale_sks_cluster_test.go
```diff
@@ -134,6 +135,7 @@ resource "exoscale_sks_cluster" "test" {
   feature_gates = []
 
   enable_kube_proxy = true
+  create_default_security_group = true
```
nerzhul, line 138: regarding the test update, it seems we don't test the setup, could we test it ?
Do we always have a test with it setup to true and one with false ?

quentinalbertone (approved):

exoscale/resource_exoscale_sks_cluster_test.go
quentinalbertone: it seems to me that Philippe checks that the value is true in the state. I think that a test with and without would test the API and not the terraform provider

kobajagi (approved):

Issue #534, opened 2026-05-20
[Bug]: nil panic after manually deleting instance


imiric:
### Current Behavior

While looking into [#453](https://github.com/exoscale/terraform-provider-exoscale/issues/453), I noticed that when a compute instance is created by the provider, and then manually removed via the CLI, portal, etc., running `terraform apply` again fails with:

<details>
  <summary>Expand</summary>

```
exoscale_security_group.my_security_group: Refreshing state... [id=d90eaf68-6eaa-49a4-b198-9c8b6637bcc9]
exoscale_private_network.my_managed_private_network: Refreshing state... [id=f9199273-1b9b-4f37-bd7d-e209119b1df5]
exoscale_elastic_ip.my_elastic_ip: Refreshing state... [id=117e9e95-f951-48fe-a044-a64e64ec12c5]
exoscale_security_group_rule.my_security_group_rule: Refreshing state... [id=07f254ae-76a1-4541-a03b-9295a54cb4f9]
exoscale_compute_instance.this: Refreshing state... [id=17caf8df-8295-4cf1-a42a-40c8c3ff1076]

Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Plugin did not respond
│
│ The plugin encountered an error, and failed to respond to the plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more details.

Stack trace from the terraform-provider-exoscale_vdev plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x153a895]

goroutine 162 [running]:
github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance.rApply({0x20488b0, 0x23ec267f6820}, 0x23ec267a2680, 0x23ec2690b320, 0x23ec262db800)
	github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance/resource.go:1193 +0x13b5
github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance.rRead({0x2048920, 0x23ec2612c000}, 0x23ec2690b320, {0x1cd1ca0, 0x23ec2620ede0})
	github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance/resource.go:515 +0x727
[...]

PR #535, opened 2026-05-20
fix(compute_instance): correctly update state if instance is manually removed

imiric:
# Description

Quick hotfix for [#534][1].

This should probably have a lower level fix instead, so opening it up for discussion.

[1]: https://github.com/exoscale/terraform-provider-exoscale/issues/534

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK

## Testing

Manually confirmed the panic no longer happens after running `terraform apply`.


imiric:
This was caused by a regression on the public API, which has been fixed. Closing.

PR #538, opened 2026-05-29
Resource compute_instance: clearify the default state value

kobajagi:
# Description

As stated in https://github.com/exoscale/terraform-provider-exoscale/pull/475, `state` has no default value, from terraform POV at least. However API does default to starting instance. We clarify this in the documentation.

Additionally makes doc generator work with opentofu.

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+IMPROVEMENTS:
+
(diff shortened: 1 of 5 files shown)


imiric (approved):
You might want to add reviewers next time 🙂

PhilippeChepy (approved):

PR #542, opened 2026-06-02
Sc 182636/upgrade private network

quentinalbertone:
# Description
Migrate private network resource and data source to tf-framework library as well as setup the the new folder structure and rewrote tests.

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

I tested the CRUD of a resource and datasource as well as using the actual version for the creation and the new version to update and delete + acc test
``` Bash
$> cd pkg/resources/private_network
$> TF_ACC=1 go test ./...
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/private_network	11.567s
```


quentinalbertone:
Acceptances test are red because of the sks oidc http call

pkg/resources/private_network/datasource.go
```diff
@@ -0,0 +1,235 @@
...
+			"description": schema.StringAttribute{
+				Description:         "The private network description.",
+				MarkdownDescription: "The private network description.",
+				Optional:            true,
+			},
+			"labels": schema.MapAttribute{
+				Description:         "A map of key/value labels.",
+				MarkdownDescription: "A map of key/value labels.",
+				ElementType:         types.StringType,
+				Optional:            true,
```
kobajagi: Should be computed though? We don't allow setting the value in data source.

pkg/resources/private_network/datasource.go
```diff
@@ -0,0 +1,235 @@
...
+				MarkdownDescription: "The Exoscale [Zone](https://www.exoscale.com/datacenters/) name.",
+				Required:            true,
+				Validators: []validator.String{
+					stringvalidator.OneOf(config.Zones...),
+				},
+			},
+			"description": schema.StringAttribute{
+				Description:         "The private network description.",
+				MarkdownDescription: "The private network description.",
+				Optional:            true,
```
kobajagi: Same question as bellow.

pkg/resources/private_network/datasource.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (d *DataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: markdownDescriptionDatasource,
+
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description:         "The private network ID to match (conflicts with `name`).",
+				MarkdownDescription: "The private network ID to match (conflicts with `name`).",
+				Optional:            true,
```
kobajagi, line 60: I think also must be tagged as `Computed`.
Because if `name` is specified we should read `id` from API and update it in state (and vice versa).

pkg/resources/private_network/datasource.go
```diff
@@ -0,0 +1,235 @@
...
+	resp.TypeName = req.ProviderTypeName + "_private_network"
+}
+
+func (d *DataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: markdownDescriptionDatasource,
+
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description:         "The private network ID to match (conflicts with `name`).",
```
kobajagi:
```suggestion
				Description:         "The private network ID to match (conflicts with 'name').",
```
No markdown here 😉

kobajagi: Also elsewhere, i will not mark individually. 
FYI `Description` is printed by terraform CLI in various places, `MarkdownDescription` is used by documentation generator (and is converted to HTML where markdown syntax makes sense).

pkg/resources/private_network/datasource.go
```diff
@@ -0,0 +1,235 @@
...
+		resp.Diagnostics.AddError(
+			"unable to change exoscale client zone",
+			err.Error(),
+		)
+		return
+	}
+
+	var privateNetwork exoscale.PrivateNetwork
+	switch {
+	case !(state.Name.IsNull() || state.Name.IsUnknown()): //nolint:staticcheck // in this case De Morgan's law is more complex to read
```
kobajagi: This condition makes me dizzy 🥲 
You only need one check, here is the explanation (depends on whether you add `Computed` attribute):
https://developer.hashicorp.com/terraform/plugin/framework/handling-data/accessing-values#when-can-a-value-be-unknown-or-null

pkg/resources/private_network/datasource.go
```diff
@@ -0,0 +1,235 @@
...
+			resp.Diagnostics.AddError(
+				"API returned an error while fetching private networks",
+				err.Error(),
+			)
+			return
+		}
+		privateNetwork, err = privateNetworks.FindPrivateNetwork(state.Name.ValueString())
+		for i := range privateNetworks.PrivateNetworks {
+			fmt.Println("->", privateNetworks.PrivateNetworks[i].Name)
+		}
```
kobajagi: Debug logs

pkg/resources/private_network/datasource.go
```diff
@@ -0,0 +1,235 @@
...
+		privateNetwork = *network
+	default: // validation must prevents this, exit as a safe guard
+		resp.Diagnostics.AddError("missing values", "name and id are missing")
+		return
+	}
+
+	state = DataSourceModel{
+		ID:          types.StringValue(privateNetwork.ID.String()),
+		Name:        types.StringValue(privateNetwork.Name),
```
kobajagi, line 204: Requires `Computed: true`, otherwise you must leave empty one attribute that is not specified in config.

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+// Metadata specifies resource name.
+func (r *Resource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_private_network"
+}
+
+func (r *Resource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description:         "Manage Exoscale Private Networks.",
+		MarkdownDescription: markdownDescriptionResource,
+		Version:             0,
```
kobajagi: Zero is default, you can skip.

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+				MarkdownDescription: "A free-form text describing the network.",
+			},
+			"labels": schema.MapAttribute{
+				Description:         "A map of key/value labels.",
+				MarkdownDescription: "A map of key/value labels.",
+				ElementType:         types.StringType,
+				Optional:            true,
+			},
+			"zone": schema.StringAttribute{
+				Description:         "❗ The Exoscale [Zone](https://www.exoscale.com/datacenters/) name.",
```
kobajagi: No markdown here

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+	zone := state.Zone.ValueString()
+	if zone == "" {
+		tflog.Info(
+			ctx,
+			"private network has no zone, deleting from state to report drift",
+			map[string]any{},
+		)
+		resp.State.RemoveResource(ctx)
+		return
+	} else if !slices.Contains(config.Zones, zone) {
```
kobajagi: Already handled on line 87

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+	zone := plan.Zone.ValueString()
+	if zone == "" {
+		tflog.Info(
+			ctx,
+			"private network has no zone, deleting from state to report drift",
+			map[string]any{},
+		)
+		resp.State.RemoveResource(ctx)
+		return
+	} else if !slices.Contains(config.Zones, zone) {
```
kobajagi: line 87

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+}
+
+func (r *Resource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	var state ResourceModel
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	timeout, diags := state.Timeouts.Read(ctx, config.DefaultTimeout)
```
kobajagi:
```suggestion
	timeout, diags := state.Timeouts.Delete(ctx, config.DefaultTimeout)
```

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	if state.ID.ValueString() == "" {
+		tflog.Info(
+			ctx,
+			"private network has no ID, deleting from state to report drift",
+			map[string]any{},
+		)
+		resp.State.RemoveResource(ctx)
```
kobajagi, line 410: I think it is removed implicitly at the end of `Delete` and you can just `return`.

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+	if zone == "" {
+		tflog.Info(
+			ctx,
+			"private network has no zone, deleting from state to report drift",
+			map[string]any{},
+		)
+		resp.State.RemoveResource(ctx)
+		return
+	} else if !slices.Contains(config.Zones, zone) {
+		resp.Diagnostics.AddError("invalid value", "zone must be a valid exoscale zone")
```
kobajagi: line 87

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+	zone := idParts[1]
+	if zone == "" {
+		tflog.Info(
+			ctx,
+			"private network has no zone, deleting from state to report drift",
+			map[string]any{},
+		)
+		resp.State.RemoveResource(ctx)
+		return
+	} else if !slices.Contains(config.Zones, zone) {
```
kobajagi, line 522: line 87

pkg/resources/private_network/resources.go
```diff
@@ -0,0 +1,546 @@
...
+	var t timeouts.Value
+	resp.Diagnostics.Append(resp.State.GetAttribute(ctx, path.Root("timeouts"), &t)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &ResourceModel{
+		ID:       types.StringValue(id.String()),
+		Zone:     types.StringValue(zone),
+		Labels:   types.MapNull(types.StringType),
```
kobajagi, line 536: Maybe fine without? Or was there issue with subsequent read? 🤔

pkg/resources/private_network/resources.go
quentinalbertone: I'm not sure and didn't find a lot of resources about this behavior so I prefer to explicitly let this one line. This behavior is mandatory for crossplane

pkg/resources/private_network/resources.go
quentinalbertone: This is the import function, the validator doesn't apply here

pkg/resources/private_network/resources.go
quentinalbertone: Without it I have this error
```
╷
│ Error: Value Conversion Error
│ 
│ An unexpected error was encountered while verifying an attribute value matched its expected type to prevent
│ unexpected behavior or panics. This is always an error in the provider. Please report the following to the
│ provider developer:
│ 
│ Expected framework type from provider logic: types.MapType[basetypes.StringType] / underlying type:
│ tftypes.Map[tftypes.String]
│ Received framework type from provider logic: types.MapType[!!! MISSING TYPE !!!] / underlying type:
│ tftypes.Map[tftypes.DynamicPseudoType]
│ Path: labels
```

kobajagi: I think `SetAttribute` instead of `Set` works better here. I had the same issue with other resources as well and wanted to address it. It's fine as is though, we can optimize it later.

kobajagi (approved):

pierre-emmanuelJ (approved):

PR #545, opened 2026-06-04
test: local-creds helper for acceptance runs

natalie-o-perret:
# Description

Local helper for running acceptance tests against real infra. Mirrors exoscale/cli#837. Reads `~/.config/exoscale/exoscale.toml`, sets the creds via `t.Setenv`, defaults `TF_ACC=1` (pre-set is respected).

```sh
go test -tags=local_integration -run TestDatabaseLocal \
  -account=owner-production -timeout 60m ./pkg/resources/database/
```

- `pkg/testutils/localcreds.go`: `LoadLocalCreds(t, account)`.
- 9 `pkg/resources/<pkg>/local_test.go` (build `local_integration`): `TestXxxLocal` per package, wraps the existing `TestXxx`. Legacy `exoscale/` left out, no single entry.
- `README.md`: short note in Contributing pointing at the helper, so new contributors know they don't need to export creds by hand.

Build, vet, gofmt, unit tests clean. Smoke tested with a fake `exoscale.toml`, env vars land correctly, `TF_ACC` default and override both work.

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK

## Testing

`make test` passes. No real-infra run, no creds in this env.

### DIFF
--- a/pkg/testutils/localcreds.go
+++ b/pkg/testutils/localcreds.go
@@ -2,26 +2,28 @@ package testutils
 
 import (
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 )
 
 // DefaultLocalAccount is the substring matched against the account name in
-// ~/.config/exoscale/exoscale.toml when -account is not provided to `go test`.
+// the local exoscale.toml when -account is not provided to `go test`.
 const DefaultLocalAccount = "owner-production"
 
-// LoadLocalCreds reads an Exoscale account block from
-// $HOME/.config/exoscale/exoscale.toml whose name contains `account` and
-// exports its credentials to the test process via t.Setenv. Mirrors the helper
-// in exoscale/cli (PR #837) so the same exoscale.toml file works for both the
-// CLI and the provider's local acceptance runs.
+// LoadLocalCreds reads an Exoscale account block from the local exoscale.toml
+// whose name contains `account` and exports its credentials to the test
+// process via t.Setenv. The config file is resolved in the same order as the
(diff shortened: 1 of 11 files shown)


natalie-o-perret:
[SC-183040]

kobajagi (approved):
Legit 👍

pkg/testutils/localcreds.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if account == "" {
+		account = DefaultLocalAccount
+	}
+
+	if os.Getenv("TF_ACC") == "" {
+		t.Setenv("TF_ACC", "1")
+	}
+
+	path := os.ExpandEnv("$HOME/.config/exoscale/exoscale.toml")
```
kobajagi: Might not work on MacOS, but that is not my problem `¯\_(ツ)_/¯`
https://github.com/exoscale/cli/blob/master/cmd/root.go#L230

natalie-o-perret: Fair, will try to tackle the x platform bits.

natalie-o-perret: Pushed. Switched to `os.UserConfigDir()` so the path matches the CLI's lookup on every platform (Linux, macOS and Windows), and honours `$EXOSCALE_CONFIG` now. 

Wdyt?

kobajagi (approved):

pkg/testutils/localcreds.go
kobajagi: All good, thanks!

PR #546, opened 2026-06-08
fix(block_storage): handle BS volume deletion in instance update

imiric:
# Description

Previously, if a block storage volume was deleted and detached at the same time, the detachment attempt during instance update would fail with:
```
Error: failed to detach block storage: DetachBlockStorageVolume: http response: Not Found: Blockstorage Volume not found
```

This is because the BS volume was already deleted in `ResourceVolume.Delete`.

The error was harmless since the volume would be deleted and detached anyways, and a subsequent `apply` would succeed, but it interrupted the process, and would cause confusion.

The fix simply checks for `v3.ErrNotFound`, allowing the instance update to continue.

Fixes internal support ticket 1167622

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK [test passes in isolation]

## Testing

1. Run the new acceptance test without the fix: produces the error, and the test fails.

2. Apply the fix.

3. Run the new acceptance test with fix applied: no error, and the test passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ IMPROVEMENTS:
 
 - Removed deprecated and unused `delay` provider attribute #539
 
+BUG FIXES:
+
+- Block storage: handle BS volume deletion in instance update #546
+
 DEPENDENCIES:
 
 - upgraded `golang/x/net` and related dependencies #540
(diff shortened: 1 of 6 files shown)


kobajagi (approved):

natalie-o-perret (approved):
👍

PR #532, opened 2026-05-18
fix(template): fall back to private visibility when name lookup returns not found

natalie-o-perret:
## Summary

Fixes #523

When `visibility` defaults to `"public"`, the list API does not return user-registered (custom) templates even when those templates show `visibility = "public"` on the template object. Fetching by ID works fine since it bypasses the visibility filter.

When a name lookup under `"public"` returns not found, the datasource now falls back to searching `"private"` before returning an error.

## Tests

```
--- PASS: TestAccDataSourceTemplate (2.02s)
ok      github.com/exoscale/terraform-provider-exoscale/exoscale        2.038s
```

PR #531, opened 2026-05-18
fix(security-group-rule): allow portless protocols (ESP, AH, GRE, IPIP)

natalie-o-perret:
## Summary

Fixes #529

`ESP`, `AH`, `GRE`, and `IPIP` protocols do not use ports or ICMP codes. The resource required `start_port`/`end_port` or `icmp_type`/`icmp_code` for all protocols, rejecting portless rules with a misleading error.

## Tests

```
--- PASS: TestSecurityGroup (29.07s)
ok      github.com/exoscale/terraform-provider-exoscale/pkg/resources/security_group    29.085s
```

PR #528, opened 2026-05-11
feat(dbaas): add exoscale_dbaas_valkey_user resource

natalie-o-perret:
Closes #478.

Every other DBaaS type has a user resource except Valkey. 

This adds `exoscale_dbaas_valkey_user`.

The egoscale v3 SDK already had all the necessary API methods (`CreateDBAASValkeyUser`, `DeleteDBAASValkeyUser`, `RevealDBAASValkeyUserPassword`), so this is mostly wiring things together following the same pattern as the OpenSearch user resource.

Tested against real infra (`bg-sof-1`): create, read, import, destroy all pass.

```
$ TF_ACC=1 go test -v -count=1 -timeout 30m -run TestDatabase/ResourceValkeyUser ./pkg/resources/database/
=== RUN   TestDatabase
=== PAUSE TestDatabase
=== CONT  TestDatabase
=== RUN   TestDatabase/ResourceValkeyUser
=== PAUSE TestDatabase/ResourceValkeyUser
=== CONT  TestDatabase/ResourceValkeyUser
--- PASS: TestDatabase (0.00s)
    --- PASS: TestDatabase/ResourceValkeyUser (153.42s)
PASS
ok      github.com/exoscale/terraform-provider-exoscale/pkg/resources/database 153.428s
```

PR #518, opened 2026-04-21
fix(compute_instance): suppress user_data diffs for pre-encoded input

natalie-o-perret:
Fixes #450

## Problem

When `user_data` is provided pre-encoded as base64 or base64+gzip, the provider passes it through unchanged on write but always decodes it on read. This causes a perpetual diff on every `terraform plan` even when nothing has changed.

## Changes

- The fix adds a `DiffSuppressFunc` that normalizes both sides (base64 decode + optional gzip decompress) before comparing, making reads and writes symmetric.
- Both `exoscale_compute_instance` and `exoscale_instance_pool` are affected and fixed here.

### Integration Tests

```
$ TF_ACC=1 EXOSCALE_API_KEY=... EXOSCALE_API_SECRET=... \
    go test -timeout=90m -tags=testacc -v -parallel=3 -count=1 \
    -run TestInstance/Resource ./pkg/resources/instance/
=== RUN   TestInstance
=== PAUSE TestInstance
=== CONT  TestInstance
=== RUN   TestInstance/Resource
=== PAUSE TestInstance/Resource
=== CONT  TestInstance/Resource
--- PASS: TestInstance (0.00s)
    --- PASS: TestInstance/Resource (551.76s)
PASS
ok      github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance  551.774s

$ TF_ACC=1 EXOSCALE_API_KEY=... EXOSCALE_API_SECRET=... \
    go test -timeout=90m -tags=testacc -v -parallel=3 -count=1 \
    -run TestInstancePool/Resource ./pkg/resources/instance_pool/
=== RUN   TestInstancePool
=== PAUSE TestInstancePool
=== CONT  TestInstancePool
=== RUN   TestInstancePool/Resource
=== PAUSE TestInstancePool/Resource
=== CONT  TestInstancePool/Resource
--- PASS: TestInstancePool (0.00s)
    --- PASS: TestInstancePool/Resource (92.89s)
PASS
ok      github.com/exoscale/terraform-provider-exoscale/pkg/resources/instance_pool    92.898s
```

### Unit Tests

```
$ go test ./pkg/utils/ -run TestSuppressUserDataDiff -v
=== RUN   TestSuppressUserDataDiff
=== PAUSE TestSuppressUserDataDiff
=== CONT  TestSuppressUserDataDiff
=== RUN   TestSuppressUserDataDiff/identical_plain_text
=== PAUSE TestSuppressUserDataDiff/identical_plain_text
[...]

PR #547, opened 2026-06-08
chore(release): add unreleased changelog entries

natalie-o-perret:
## Summary

Backfill the `Unreleased` block in `CHANGELOG.md` with natalie's PRs merged since v0.69.2 (2026-05-05) that were missing.

## Related issue

None.

## Changes

Add the following to the existing `## Unreleased` block in `CHANGELOG.md`:

### IMPROVEMENTS
- test: local-creds helper for acceptance runs #545

### BUG FIXES
- fix(compute_instance): suppress user_data diffs for pre-encoded input #518
- fix(security-group-rule): allow portless protocols (ESP, AH, GRE, IPIP) #531
- fix(template): fall back to private visibility when name lookup returns not found #532

### FEATURES
- feat(dbaas): add exoscale_dbaas_valkey_user resource #528

## Checklist

* [ ] Acceptance tests OK (not applicable, docs only)

## Notes for reviewers

Entries for other merged PRs since 0.69.2 (#533 by @imiric) are missing. Not added here per scope; left for the relevant author or release manager.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ DEPENDENCIES:
 
 FEATURES:
 
-- feat(dbaas): add exoscale_dbaas_valkey_user resource #528
+- `dbaas`: add exoscale_dbaas_valkey_user resource #528
 
 ## 0.69.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ IMPROVEMENTS:
 
 BUG FIXES:
 
-- Block storage: handle BS volume deletion in instance update #546
+- `block_storage_volume`: handle BS volume deletion in instance update #546
 - `compute_instance`: suppress user_data diffs for pre-encoded input #518
 - `security_group_rule`: allow portless protocols (ESP, AH, GRE, IPIP) #531
 - `template`: fall back to private visibility when name lookup returns not found #532
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,9 @@ IMPROVEMENTS:
 BUG FIXES:
 
 - Block storage: handle BS volume deletion in instance update #546
-- fix(compute_instance): suppress user_data diffs for pre-encoded input #518
-- fix(security-group-rule): allow portless protocols (ESP, AH, GRE, IPIP) #531
-- fix(template): fall back to private visibility when name lookup returns not found #532
+- `compute_instance`: suppress user_data diffs for pre-encoded input #518
+- `security_group_rule`: allow portless protocols (ESP, AH, GRE, IPIP) #531
+- `template`: fall back to private visibility when name lookup returns not found #532
 
 DEPENDENCIES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@
 IMPROVEMENTS:
 
 - Removed deprecated and unused `delay` provider attribute #539
-- test: local-creds helper for acceptance runs #545
 
 BUG FIXES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,22 @@
 IMPROVEMENTS:
 
 - Removed deprecated and unused `delay` provider attribute #539
+- test: local-creds helper for acceptance runs #545
 
 BUG FIXES:
 
 - Block storage: handle BS volume deletion in instance update #546
+- fix(compute_instance): suppress user_data diffs for pre-encoded input #518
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,15 +5,23 @@
...
 
 - Removed deprecated and unused `delay` provider attribute #539
+- test: local-creds helper for acceptance runs #545
 
 BUG FIXES:
 
 - Block storage: handle BS volume deletion in instance update #546
+- fix(compute_instance): suppress user_data diffs for pre-encoded input #518
+- fix(security-group-rule): allow portless protocols (ESP, AH, GRE, IPIP) #531
+- fix(template): fall back to private visibility when name lookup returns not found #532
```
imiric: I would drop the type since it's already clear from the section, and make the scope a bit more user friendly, so:

```suggestion
- Instance: suppress user_data diffs for pre-encoded input #518
- Security group rule: allow portless protocols (ESP, AH, GRE, IPIP) #531
- Template: fall back to private visibility when name lookup returns not found #532
```

Although it would also make sense to use the suffix from the actual TF resource name, so:
```suggestion
- `compute_instance`: suppress user_data diffs for pre-encoded input #518
- `security_group_rule`: allow portless protocols (ESP, AH, GRE, IPIP) #531
- `template`: fall back to private visibility when name lookup returns not found #532
```

... could also work.

TBH we don't follow a strict convention about this, but maybe we should. WDYT?

CHANGELOG.md
```diff
@@ -5,15 +5,23 @@
 IMPROVEMENTS:
 
 - Removed deprecated and unused `delay` provider attribute #539
+- test: local-creds helper for acceptance runs #545
```
imiric: I'm not sure whether it makes sense to mention any dev or test-related changes. This is why I didn't include #533 here.

My rationale is that the changelog should mention user-facing changes, and ideally be directed towards the user (there's an argument to be made that commit messages shouldn't be used in a changelog directly for this reason either). So any changes that don't strictly impact the user are just noise from the perspective of a user IMO.

But there is no universal rule for this, so happy to hear other thoughts.

imiric (commented):
Thanks! Just some minor feedback, mostly to align on a single convention.

natalie-o-perret:
> Thanks! Just some minor feedback, mostly to align on a single convention.

So my stance (which isn't everybody's stance) we should include everything, ideally (I know, I like this word), in the future we could lean towards release-please, so that if forces folks (myself included) to have meaningful commits from which the rest is derived (fwiw, we can tweak the conv comm/lint/release-please, so that it doesn't bloat changelog with unnecessary details, like e.g., ci-centric contributions)

CHANGELOG.md
natalie-o-perret: Agreed on #545, will drop, given how things are done today.

On the rewording, I'd keep commit-style for now since it matches the 0.69.2 block above (e.g. sks: add Kubelet Image GC..., Fix SG resource schema version).

I'd rather do it as one pass over the whole CHANGELOG + CI/CD than piecemeal per PR. 

Will create a note so we can discuss it internally.

CHANGELOG.md
```diff
@@ -9,11 +9,18 @@ IMPROVEMENTS:
 BUG FIXES:
 
 - Block storage: handle BS volume deletion in instance update #546
```
imiric: Unrelated to your change, but so that we're at least consistent within a single release:

```suggestion
- `block_storage_volume`: handle BS volume deletion in instance update #546
```

nerzhul (approved):

CHANGELOG.md
```diff
@@ -9,11 +9,18 @@ IMPROVEMENTS:
...
+- `security_group_rule`: allow portless protocols (ESP, AH, GRE, IPIP) #531
+- `template`: fall back to private visibility when name lookup returns not found #532
 
 DEPENDENCIES:
 
 - upgraded `golang/x/net` and related dependencies #540
 
+FEATURES:
+
+- feat(dbaas): add exoscale_dbaas_valkey_user resource #528
```
imiric:
```suggestion
- `dbaas`: add exoscale_dbaas_valkey_user resource #528
```

imiric (approved):
Thanks! 🙇

CHANGELOG.md
natalie-o-perret: Ta :bowing_woman:

CHANGELOG.md
natalie-o-perret: Ta :bowing_woman:

PR #549, opened 2026-06-10
fix: tolerate missing instance during detach

imiric:
# Description

`detachMatchingResource` lists instances and then fetches each one to check for a matching attachment before deleting a resource. If an instance is deleted between the `ListInstances` and `GetInstance` calls (e.g. by a concurrent operation), `GetInstance` returns `ErrNotFound` which previously aborted the entire detach operation and left dangling resources.

A vanished instance can no longer hold the attachment, so skip it instead of failing. Other `GetInstance` errors are still propagated.

This is an attempt at fixing the [flaky failure of `TestAccDataSourceElasticIP`][1]:
```
datasource_exoscale_elastic_ip_test.go:125: Error running post-test destroy, there may be dangling resources: exit status 1

    Error: GetInstance: http response: Not Found: Instance not found.
```

I wasn't able to reproduce this locally (running the acceptance tests against my prod account), but this is the only scenario I can think of where this could happen.

Note that `detachMatchingResource` is also called by `resourcePrivateNetworkDelete`, but I think the same logic applies in that case as well.

[1]: https://github.com/exoscale/terraform-provider-exoscale/actions/runs/27135434233/job/80352641300

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK

## Testing

I think it's fixed :tada: Or, at least, `TestAccDataSourceElasticIP` didn't fail in [this run](https://github.com/exoscale/terraform-provider-exoscale/actions/runs/27278126311/job/80569409971).

There are still two failures left: `TestAccResourceSKSClusterSKSClusterWithAudit` and `TestInstance/Resource`.

We could run it a few more times before merging, if you'd like to have more confidence. Though we can also easily revert it if it fails on `master`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ BUG FIXES:
 - `compute_instance`: suppress user_data diffs for pre-encoded input #518
 - `security_group_rule`: allow portless protocols (ESP, AH, GRE, IPIP) #531
 - `template`: fall back to private visibility when name lookup returns not found #532
+- `elastic_ip`: tolerate missing instance during detach #549
 
 DEPENDENCIES:
 
--- a/exoscale/util.go
+++ b/exoscale/util.go
@@ -2,6 +2,7 @@ package exoscale
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"regexp"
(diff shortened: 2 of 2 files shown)


natalie-o-perret (approved):
🙏 Thanks!

kobajagi (approved):

PR #551, opened 2026-06-11
fix(instance): check instance state before stop/start

imiric:
# Description

Previously `instance.rUpdate` would attempt to stop an instance regardless of its current state, causing errors like:
```
StopInstance: http response: Bad Request: Only running instances can be stopped.
```

This was reproducible by our [`TestInstance/Resource` acceptance test][1].

The fix checks that the instance is actually running before attempting to stop it. And, for good measure, does the same when attempting to start it later, even though we haven't run into any issues with that call.

The second `GetInstance` to refresh the state isn't technically required (I've observed the instance is stopped after the first call), but I added it for additional safety in case of drift, since there are many intermediate operations.

Sidenote: this function should really be refactored, since it's very long and difficult to follow. The `nolint:gocyclo` is a hint. But I left that for later.

[1]: https://github.com/exoscale/terraform-provider-exoscale/actions/runs/27335561537/job/80763036826#logs

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

`TF_ACC=1 go test -v -count=1 -timeout 15m ./pkg/resources/instance` passes locally, whereas it didn't before this change.

### DIFF
--- a/pkg/resources/instance/resource.go
+++ b/pkg/resources/instance/resource.go
@@ -860,11 +860,21 @@ func rUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnos
 			return diag.Errorf("unable to scale down the disk size, use size > %v", instance.DiskSize)
 		}
 
-		// Compute instance scaling/disk resizing/TPM enabling API operations requires the instance to be stopped.
-		if d.Get(AttrState) == "stopped" ||
-			d.HasChange(AttrDiskSize) ||
-			d.HasChange(AttrType) ||
-			(d.HasChange(AttrEnableTPM) && d.Get(AttrEnableTPM).(bool)) {
+		// Refresh instance state, since it could have changed since when we
+		// first requested it above.
+		instance, err = client.GetInstance(ctx, v3.UUID(d.Id()))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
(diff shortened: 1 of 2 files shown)


natalie-o-perret (approved):

pkg/resources/instance/resource.go
```diff
@@ -918,15 +927,16 @@ func rUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnos
 			}
 		}
 
-		if d.Get(AttrState) == "running" {
+		shouldStart := d.Get(AttrState) == string(v3.InstanceStateRunning) &&
+			(shouldStop || instance.State != v3.InstanceStateRunning)
```
kobajagi, line 931: Can you add a comment on top explaining what flows are covered here?
It is very much non-obvious and took me few minutes to mind map it.

kobajagi (approved):

pkg/resources/instance/resource.go
imiric: You mean for this particular check before starting an instance, or for the case above when stopping it?

I didn't see any failures for this start call, but it's just a defensive check similar to the stop one. But then again, I suppose if the instance is running at this point, any of the previous calls would fail anyway. Is this what you're referring to?

I can remove this particular change, if you prefer.

PR #552, opened 2026-06-11
ci: split unit/acceptance test workflows

imiric:
# Description

Acceptance tests are very resource intensive to run, and triggering them on every commit on every branch is too permissive. Especially since they're run sequentially and block other test runs.

E.g. sometimes I push a commit and notice I forgot something, so amending and force pushing means having to wait another ~30m... And I can't interrupt the workflow since it will likely leave dangling resources.

So this splits the previous `tests` workflow into `test-unit` and `test-acc`. Unit tests are still run on every commit, since they're quick, cheap, and can run in parallel. Acceptance tests only run on pull requests to `master`, and on direct commits on `master`, and are serialized as before.

> [!NOTE]
> This was drafted with the assistance of LLMs, so I'd like to test it works as expected, especially the `workflow_dispatch` case.

## Checklist
(For exoscale contributors)

* [ ] Changelog updated (under *Unreleased* block) [not worthy]
* [ ] Acceptance tests OK

## Testing

Looks fine in this PR, but we should test it after merging.


imiric:
Welp, I can't actually test `workflow_dispatch` on the new workflow, since "Run workflow" only becomes available if the file is on the default branch. And I can't run the old one since it was renamed:


So... I propose we review this manually, and I'll test it once it's merged.

kobajagi (approved):

natalie-o-perret (approved):

imiric:
Manual workflow run with `"^Test.*Database.*"` [worked fine](https://github.com/exoscale/terraform-provider-exoscale/actions/runs/27355325971/job/80827955627).

PR #555, opened 2026-06-15
fix(dbaas_pg_database): wait for new database to appear in service list

natalie-o-perret:
# Description

After `CreateDBAASPGDatabase` returns success, the service's `Databases` list can lag the operation by a few seconds. The previous code did a single `GetDBAASServicePG` call and bailed on the first miss with `"Unable to find newly created database for the service"`, which made `TestDatabase/ResourcePg` flake whenever the control plane was slow to propagate.

This swaps the single read for a poll using the existing `waitForDBAASServiceReadyForFn` helper, with a predicate that returns true once the new database appears in the service's `Databases` list. Same shape as the MySQL database resource, which already does this correctly.

## Checklist
(For exoscale contributors)

- [x] Changelog updated (under *Unreleased* block)
- [ ] Acceptance tests OK
- [ ] For a new resource, datasource or new attributes: acceptance test added/updated

## Testing

- `go build ./pkg/resources/database/...` clean
- `go vet ./pkg/resources/database/...` clean
- Acceptance tests for `TestDatabase/ResourcePg` will be exercised by CI on this PR

> [!NOTE]
> AI-assisted for the diff and the changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+BUG FIXES:
+
+- `dbaas_pg_database`: wait for newly created database to appear in service list, fixing intermittent "Unable to find newly created database for the service" error on create
+
 IMPROVEMENTS:
 
 - Resource compute_instance: clearify the default state value #538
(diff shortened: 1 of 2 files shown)


natalie-o-perret:
[SC-184072]

PR #560, opened 2026-06-23
feat(ci): enable Go test caching in GitHub Actions workflows

pierre-emmanuelJ:
Enable Go's built-in test result caching across CI runs by

This allows unchanged packages to skip re-execution on subsequent runs,
dramatically reducing acceptance test duration and avoiding flaky test
failures for code that has not changed.

### DIFF
--- a/.github/workflows/test-acc.yml
+++ b/.github/workflows/test-acc.yml
@@ -51,4 +51,4 @@ jobs:
         env:
           EXOSCALE_API_KEY: ${{ secrets.EXOSCALE_API_KEY }}
           EXOSCALE_API_SECRET: ${{ secrets.EXOSCALE_API_SECRET }}
-          EXTRA_ARGS: ${{ github.event.inputs.test_run_pattern != '' && format('-parallel=3 -failfast -run={0}', github.event.inputs.test_run_pattern) || '' }}
+          EXTRA_ARGS: ${{ inputs.test_run_pattern != '' && format('-parallel=3 -count=1 -failfast -run={0}', inputs.test_run_pattern) || '' }}
--- a/.github/workflows/test-acc.yml
+++ b/.github/workflows/test-acc.yml
@@ -34,10 +34,12 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - run: make go.mk
-      - uses: ./go.mk/.github/actions/setup
+      - name: Set up Go
+        uses: actions/setup-go@v6
         with:
           go-version-file: 'go.mod'
-          cache: 'true'
+          cache: true
+          cache-dependency-path: 'go.sum'
       - uses: ./go.mk/.github/actions/pre-check
       - name: Run acceptance tests
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test-acc.yml
```diff
@@ -36,14 +36,23 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-go-${{ hashFiles('go.sum') }}-
+            ${{ runner.os }}-go-
       - name: Run acceptance tests
         run: |
           if [ -n "${EXTRA_ARGS}" ]; then
             make test-acc EXTRA_ARGS="${EXTRA_ARGS}"
           else
-            make test-acc
+            make test-acc EXTRA_ARGS="-parallel=3 -failfast"
```
quentinalbertone: **thought:** I think that `make test-acc` already has `-parallel=3 -failfast`

.github/workflows/test-acc.yml
```diff
@@ -36,14 +36,23 @@ jobs:
...
           fi
         env:
           EXOSCALE_API_KEY: ${{ secrets.EXOSCALE_API_KEY }}
           EXOSCALE_API_SECRET: ${{ secrets.EXOSCALE_API_SECRET }}
-          EXTRA_ARGS: ${{ inputs.test_run_pattern != '' && format('-parallel=3 -count=1 -failfast -run={0}', inputs.test_run_pattern) || '' }}
```
natalie-o-perret, line 49: In hindsight, don't we have a risk for stale pass?  🤔

pierre-emmanuelJ: We could make the cache expire every 24 h. Maybe for that, the idea could be one test cache per day and facilitate a business hour day.

.github/workflows/test-acc.yml
```diff
@@ -36,14 +36,23 @@ jobs:
       - run: make go.mk
       - uses: ./go.mk/.github/actions/setup
       - uses: ./go.mk/.github/actions/pre-check
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('go.sum') }}-${{ github.run_id }}
```
natalie-o-perret: Unless I've got that wrong, `run_id` might make every run a cache miss and would defeat the whole purpose of this PR 🤔, the `go.sum` isn't enough on its own? 🤔

pierre-emmanuelJ: yes sum only should be better 👍

.github/workflows/test-acc.yml
```diff
@@ -36,6 +36,15 @@ jobs:
...
       - uses: ./go.mk/.github/actions/pre-check
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-${{ hashFiles('go.sum') }}-
+            ${{ runner.os }}-go-
```
pierre-emmanuelJ: still need to fix this one

.github/workflows/test-acc.yml
kobajagi: I wouldn't call it stale, no code change = no new test run.
API issues that would cause subsequent test runs to fail are not immediate tf-provider dev concern.
We could keep `-count=1` in master branch only.

natalie-o-perret (approved):
🙏

quentinalbertone (approved):

PR #562, opened 2026-07-02
feat: only use security group id for private security group rule

quentinalbertone:
# Description
if you pass a security-group as source in a rule, only a uuid is needed for a private sg

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

```Bash
$> TF_ACC=1 go test -tags=testacc ./pkg/resources/security_group/... -v
=== RUN   TestSecurityGroup
=== PAUSE TestSecurityGroup
=== CONT  TestSecurityGroup
--- PASS: TestSecurityGroup (37.71s)
PASS
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/security_group	37.745s
```


pkg/resources/security_group/resource_rule.go
```diff
@@ -737,8 +736,7 @@ func (r *ResourceRule) FindRemoteRule(
 			((item.SecurityGroup == nil && rule.SecurityGroup == nil) ||
 				(item.SecurityGroup != nil &&
 					rule.SecurityGroup != nil &&
-					item.SecurityGroup.ID == rule.SecurityGroup.ID &&
-					item.SecurityGroup.Name == rule.SecurityGroup.Name)) &&
```
kobajagi, line 741: Is this correct? PR description mentions only `add-rule-to-security-group` request not needing name parameter, this matches `get-security-group` output.

quentinalbertone: Without this change the code was not working and didn't find the security group rule and I assume that if security group ids where matching the name was optional

kobajagi (approved):

quentinalbertone:
The acc test for the security group is green and the sks error comes from the sks API, so we will merge this PR

PR #585, opened 2026-09-03
VPC integration: subnets

sauterp:
# Description

We add the resource and data-source to define VPC subnets.

[sc-196385]

## Checklist
(For exoscale contributors)

* [x] Acceptance tests OK

## Testing

```shell
> TF_ACC=1 go test -tags=testacc \
./pkg/resources/vpc/... -v
=== RUN   Test_Resource_VPC
=== PAUSE Test_Resource_VPC
=== RUN   Test_Resource_VPC_Subnet
=== PAUSE Test_Resource_VPC_Subnet
=== CONT  Test_Resource_VPC
=== CONT  Test_Resource_VPC_Subnet
--- PASS: Test_Resource_VPC (5.00s)
--- PASS: Test_Resource_VPC_Subnet (7.09s)
PASS
ok  	github.com/exoscale/terraform-provider-exoscale/pkg/resources/vpc	7.123s
```

I verified the examples:
```
  # exoscale_vpc.my_vpc will be created
  + resource "exoscale_vpc" "my_vpc" {
      + default     = false
      + description = "My Virtual Private Cloud"
      + id          = (known after apply)
      + labels      = {
          + "environment" = "production"
        }
      + name        = "my-vpc"
      + zone        = "ch-gva-2"
    }

  # exoscale_vpc_subnet.my_vpc_subnet will be created
  + resource "exoscale_vpc_subnet" "my_vpc_subnet" {
      + address_family = "inet4"
      + address_space  = "private"
      + description    = "My VPC Subnet"
      + id             = (known after apply)
      + ipv4_block     = "10.0.0.0/24"
      + labels         = {
          + "environment" = "production"
        }
      + name           = "my-vpc-subnet"
      + vpc_id         = (known after apply)
      + zone           = "ch-gva-2"
    }
```

after applying:
```
❯ exo net vpc -z $ZONE subnet show my-vpc my-vpc-subnet
┼────────────────┼──────────────────────────────────────┼
│   VPC SUBNET   │                                      │
┼────────────────┼──────────────────────────────────────┼
│ ID             │ 2d7d6c21-f738-4418-9fc5-d219764b9d9e │
│ Name           │ my-vpc-subnet                        │
│ Description    │ My VPC Subnet                        │
│ Zone           │ ch-gva-2                             │
│ Created At     │ 2026-09-17 15:39:03 +0000 UTC        │
│ Address Family │ inet4                                │
│ Address Space  │ private                              │
│ IPv4 Block     │ 10.0.0.0/24                          │
│ Labels         │ production:                          │
│ Instances      │ -                                    │
┼────────────────┼──────────────────────────────────────┼

```


pkg/resources/vpc/resource_subnet.go
```diff
@@ -0,0 +1,491 @@
...
+		return
+	}
+
+	resp.Diagnostics.AddAttributeError(
+		req.Path,
+		"Changing the Subnet CIDR is not supported",
+		"Resizing a Subnet is not supported\n\n"+
+			"Current: "+req.StateValue.ValueString()+"\nPlanned: "+req.PlanValue.ValueString()+
+			"\n\nRestore the ipv4_block value in the configuration, or create a new Subnet "+
+			"with the desired CIDR.",
```
sauterp: Confirmed that it errors if the user tries to resize the subnet:
```
exoscale_vpc.my_vpc: Refreshing state... [id=e174b9c2-ff0b-4216-a2f1-13be20b63795]
exoscale_vpc_subnet.my_vpc_subnet: Refreshing state... [id=c3d7ae6e-ea6b-4dd7-8346-278e2f7812c6]

Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Changing the Subnet CIDR is not supported
│
│   with exoscale_vpc_subnet.my_vpc_subnet,
│   on resource.tf line 20, in resource "exoscale_vpc_subnet" "my_vpc_subnet":
│   20:   ipv4_block  = "10.0.0.0/16"
│
│ Resizing a Subnet is not supported
│
│ Current: 10.0.0.0/24
│ Planned: 10.0.0.0/16
│
│ Restore the ipv4_block value in the configuration, or create a new Subnet with the desired CIDR.
╵
```

sauterp: This error is needed because we don't support resizing and there is no field for the `ipv4_block` in the `update-subnet` call, so we won't get an error from the API. Without this error tf will always have an unclean plan if the user sets a different CIDR, which is confusing.

sauterp:
I'll add the Changelog line just before merging, to avoid an annoying conflict whenever somebody else merges something.

docs/resources/vpc_subnet.md
```diff
@@ -0,0 +1,97 @@
...
+
+Please refer to the [examples](https://github.com/exoscale/terraform-provider-exoscale/tree/master/examples/)
+directory for complete configuration examples.
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `ipv4_block` (String) The Subnet IPv4 CIDR (e.g. `10.0.0.0/24`). Cannot be changed after creation.
```
aureliar8: It's interesting to me that from the terraform POV this field cannot be change while the zone can be (even if it means to actually delete + recreate) 
Shouldn't we make it work like the zone where changing the ipv4_block actually deletes & recreate a subnet ?

pkg/resources/vpc/resource.go
```diff
@@ -376,6 +377,17 @@ func (r *Resource) Delete(ctx context.Context, req resource.DeleteRequest, resp
...
+// noLabels is the state value to use when the API reports no labels. It keeps
+// the prior state's representation so that a config with an explicit
+// `labels = {}` stays an empty map instead of flipping to null on every
+// refresh, which Terraform would report as permanent drift.
+func noLabels(prior types.Map) types.Map {
+	if !prior.IsNull() && !prior.IsUnknown() && len(prior.Elements()) == 0 {
+		return types.MapValueMust(types.StringType, nil)
+	}
+	return types.MapNull(types.StringType)
+}
```
aureliar8, line 389: I'm curious as to why this helper only appears in VPC. Do we behave differently than the rest of the exoscale API ?

docs/resources/vpc_subnet.md
sauterp: As discussed, `ipv4_block` changes now force replacement.

pkg/resources/vpc/resource.go
sauterp: Much of the provider is still implemented in tf SDK v2 and not the new 'framework'. And the SDK type used for labels is usually TypeMap, which doesn't distinguish between null and empty unlike framework. I'd say that's why we don't have this issue for most resource labels.
A resource implemented in 'framework' is `private_network` and I just verified, it actually has the issue I'm fixing here. If I set `labels = {}` I get "permanent drift".

sauterp: created a card: https://app.shortcut.com/exoscale/story/199669/tf-provider-private-network-resource-empty-labels-causes-permanent-drift

docs/resources/vpc.md
```diff
@@ -29,6 +29,18 @@ resource "exoscale_vpc" "my_vpc" {
...
+  zone        = local.zone
+  vpc_id      = exoscale_vpc.my_vpc.id
+  name        = "my-vpc-subnet"
+  description = "My VPC Subnet"
+  ipv4_block  = "10.0.0.0/24"
+
+  labels = {
+    environment = "production"
+  }
+}
```
quentinalbertone, line 43: **niptick:** you should not mention  `exoscale_vpc_subnet` resources in the `exoscale_vpc` doc

docs/data-sources/vpc.md
```diff
@@ -21,9 +21,19 @@ data "exoscale_vpc" "my_vpc" {
   name = "my-vpc"
 }
 
+data "exoscale_vpc_subnet" "my_vpc_subnet" {
+  zone   = "ch-gva-2"
+  vpc_id = data.exoscale_vpc.my_vpc.id
+  name   = "my-vpc-subnet"
+}
+
```
quentinalbertone, line 29: **niptick:** you should not mention  `exoscale_vpc_subnet` datasource in the `exoscale_vpc` doc

docs/resources/vpc.md
```diff
@@ -73,4 +85,10 @@ Optional:
 terraform import \
   exoscale_vpc.my_vpc \
   f81d4fae-7dec-11d0-a765-00a0c91e6bf6@ch-gva-2
+
+# An existing VPC Subnet may be imported by `<vpc-ID>@<subnet-ID>@<zone>`:
+
+terraform import \
+  exoscale_vpc_subnet.my_vpc_subnet \
+  f81d4fae-7dec-11d0-a765-00a0c91e6bf6@9ecc6b8b-73d4-4211-8ced-f7f29bb79524@ch-gva-2
```
quentinalbertone, line 93: **niptick:** you should not mention  `exoscale_vpc_subnet` resources in the `exoscale_vpc` doc

pkg/resources/vpc/resource_subnet.go
```diff
@@ -0,0 +1,455 @@
...
+				Required:            true,
+				Description:         "The Subnet name.",
+				MarkdownDescription: "The Subnet name.",
+			},
+			"description": schema.StringAttribute{
+				Optional:            true,
+				Description:         "A free-form text describing the Subnet.",
+				MarkdownDescription: "A free-form text describing the Subnet.",
+			},
+			"ipv4_block": schema.StringAttribute{
```
quentinalbertone, line 86: **suggestion** you could create a custom validator for this field and prevent api call that will eventually fail. Here is an example: https://github.com/exoscale/terraform-provider-exoscale/blob/master/pkg/resources/private_network/resources.go#L93

quentinalbertone (changes requested):

pkg/resources/vpc/resource_subnet.go
aureliar8: Could you expend as to _why_ this is a good idea ? What the upside in doing client side validation ?
Because the downside is that we'll need to ensure that two different system are always in sync (including one that we have no control over it's version bumps)

pkg/resources/vpc/resource_subnet.go
```diff
@@ -0,0 +1,455 @@
...
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				MarkdownDescription: "The ID of this resource.",
+				Computed:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"vpc_id": schema.StringAttribute{
+				Description:         "The parent VPC ID.",
```
quentinalbertone, line 58: This field force a replace if updated

```suggestion
				Description:         "❗ The parent VPC ID.",
```

quentinalbertone (changes requested):

pkg/resources/vpc/resource_subnet.go
quentinalbertone: From what I understood, you are expecting a CIDR for this field. You treat it as a string in the schema and in the create function so the user could trigger an api call to the backend with a string `1.1.1/32` or `not a cidr` (for example). I guess that this call will always fail.

It could have been detected in the client: during the plan or before the api call: for a simple vpc + subnet it's nothing but if you want to deploy a large infra an error in the plan is the best in contrary to an error during the apply.

Plus I also guess that a parse IPv4 in the client will always be sync with the parse IPv4 in the server in the backend

aureliar8: OK now I understand the difference between validation during plan and during apply. It's indeed a nicer UX if we can detect errors during a plan than during an apply 

However here, even though you are correct that _right now_ the API only accept CIDR, we have plans wto change that in the future. (You could pass other stuff, like for example a hardcoded string "public-ip" or a  resource identifier to a IP pool that you own) 

When such change happen I want to only have to update public-api, not every client-side tool that we use that tried to re-implement its own idea of validation  

Hence I this case I'd say it's better to not validate inside terraform, to be able to iterate faster (at a small cost of user experience)

quentinalbertone: ok I see what you want to do. It was a suggestion form the start, it's fine by me if you want to keep the code as it is now

pkg/resources/vpc/datasource_subnet.go
```diff
@@ -0,0 +1,223 @@
...
+		if err != nil {
+			resp.Diagnostics.AddError("unable to parse ID", err.Error())
+			return
+		}
+		got, err := client.GetSubnet(ctx, vpcID, id)
+		if err != nil {
+			resp.Diagnostics.AddError("API returned an error while fetching Subnet", err.Error())
+			return
+		}
+		subnet = *got
```
aureliar8: Can we swap to prioritize lookups by uuid rather than by name?

aureliar8 (approved):

pierre-emmanuelJ (approved):

PR #587, opened 2026-09-03
VPC integration: subnet attachments

sauterp:
# Description

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK


sauterp:
We decided that attachments will be a block on `exoscale_compute_instance` rather than a separate resource.

PR #593, opened 2026-09-17
Migrate nlb/nlb_service to framework & egoscale v3

pierre-emmanuelJ:
Moves the whole NLB family out of the plugin-SDKv2 provider into a single `pkg/resources/nlb` package on terraform-plugin-framework + egoscale v3, following the `security_group` / `block_storage` / `vpc` layout.

## Schema

Preserved attribute for attribute — `terraform providers schema -json` before/after reports **no cty type change** on any of the four schemas. Only flag changes (`optional` → `optional,computed` where a `Default` requires it).

Additive fixes: `labels` and `timeouts` on the `exoscale_nlb` data source, the swapped `port` / `healthcheck.port` descriptions, and `services` documented as IDs (it always held IDs, not names).

## Behaviour notes

- SDKv2 persists unset optional strings as `""`, the framework as `null`, so `description` / `uri` / `tls_sni` are normalised on read — otherwise pre-migration state produces a permanent diff.
- Updates send only the attributes that changed, with two exceptions. The load-balancer endpoint rejects an *empty* payload, so clearing `labels` goes through `ResetLoadBalancerField` rather than an update. The service endpoint requires `protocol` even when unchanged (it answers ``Invalid value `:HTTP/1.1` at 'protocol'`` otherwise, the quoted value being the server misreporting an absent field), so `protocol` and `strategy` are always sent.
- The healthcheck is always sent whole, which is what clears `uri`/`tls_sni` when they are dropped from config.
- `Create` and `Update` write only computed attributes; `Read` is the only place that refreshes everything, so API-side normalisation surfaces as drift instead of an inconsistent-result error.
- Looking up an NLB by name now goes through `FindLoadBalancer`, which errors on duplicate names instead of returning the first match (same as `vpc` / `security_group`).

## Known issue

Emptying `description` is not applied — see the `TODO(egoscale)` in `resource.go`. `UpdateLoadBalancer(Service)Request.Description` is a non-nullable `string` (unlike `UpdateVpcRequest.Description`, a `*string`), so an emptied value is never serialised, and `DELETE /load-balancer/{id}/description` returns 5xx. Sending an explicit `"description": ""` does clear it, so this is a one-line egoscale fix. Clearing `labels` works via `ResetLoadBalancerField`.

## Testing

Acceptance tests against `bg-sof-1`:

- `TestNLB` — create, update, clear labels, import NLB (`id@zone`) and service (`nlb/svc@zone`) — PASS
- `TestNlbService/DataSourceList` — PASS

Manual old→new migration, swapping provider binaries over the same state via `dev_overrides`, with state written by the pre-migration binary containing a two-element `services` set, a populated https healthcheck (`uri` + `tls_sni`), and the `""` empty-string case:

- `plan` with the new binary → **No changes.**
- `apply` a change onto that state → applied, re-plan → **No changes.**
- `state rm` + `import` of resources created by the old binary → **No changes.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pierre-emmanuelJ:
I hold this PR until the API spec has been updated to **nullable** `description` field in update NLB call.

pkg/resources/nlb/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
+	"github.com/exoscale/terraform-provider-exoscale/pkg/testutils"
+)
+
+func TestNlbService(t *testing.T) {
+	t.Run("DataSourceList", testListDataSource)
+}
+
+func TestNLB(t *testing.T) {
```
kobajagi, line 14: We need to merge these 2, datasource can be tested at the same time as the resource.

pkg/resources/nlb/models.go
```diff
@@ -0,0 +1,173 @@
...
+		Timeout:  types.Int64Value(hc.Timeout),
+		TLSSNI:   optionalString(hc.TlsSNI),
+		URI:      optionalString(hc.URI),
+	}
+}
+
+// optionalString returns a null value for an empty string. The SDKv2 provider
+// persisted unset optional strings as "", while the framework expects null;
+// normalising here keeps pre-migration state from producing a permanent diff.
+func optionalString(s string) types.String {
```
kobajagi: Can move to utils package.

pkg/resources/nlb/resource.go
```diff
@@ -0,0 +1,467 @@
...
+		return
+	}
+
+	nlb, err := client.GetLoadBalancer(ctx, operation.Reference.ID)
+	if err != nil {
+		resp.Diagnostics.AddError("API returned an error while fetching the created NLB", err.Error())
+		return
+	}
+
+	resp.Diagnostics.Append(applyNLB(ctx, &plan, nlb)...)
```
kobajagi: Refreshing on create (what applyNLB does) is not a good idea. Framework checks for unexpected changes in state (optional / required attributes not matching plan)  and errors out with infamous error "This is always bug in provider yadayada". Normally this doesn't happen as upstream nlb we just created matches plan, but any discrepancies (normalization of values or just a bug in the api) would throw error, instead it should allow next plan to report a drift.

kobajagi: To summarize, only set computed values in `Create` & `Update`. Only `Read` should refresh complete state (every attribute).

pkg/resources/nlb/resource.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+
+	update := !plan.Name.Equal(state.Name) ||
+		!plan.Description.Equal(state.Description) ||
+		!plan.Labels.Equal(state.Labels)
+
+	// The full desired state is sent rather than just the changed fields, to
+	// match what the egoscale v2 provider did (it mutated and re-sent the whole
+	// object) and to stay consistent with the service update, where a partial
+	// request is rejected by the API.
```
kobajagi: It does not mater how v2 handles it unless there is a state drift, which there isn't in this case. Second part is also untrue: API docs state all attributes are optional and i confirmed it with cli:
```
>>> Operation: update-load-balancer
PUT /v2/load-balancer/a0c4275a-8ba6-45ea-b028-3c88b93820de HTTP/1.1
...
{"description":"rest"}
<<< HTTP/2.0 200 OK
...
```

pkg/resources/nlb/resource.go
```diff
@@ -0,0 +1,467 @@
...
+
+	nlb, err := client.GetLoadBalancer(ctx, id)
+	if err != nil {
+		resp.Diagnostics.AddError("API returned an error while fetching the updated NLB", err.Error())
+		return
+	}
+
+	plannedDescription := plan.Description
+
+	resp.Diagnostics.Append(applyNLB(ctx, &plan, nlb)...)
```
kobajagi: See my comment on `Create`, no full refresh outside of `Read`.

pkg/resources/nlb/main_test.go
pierre-emmanuelJ: https://github.com/exoscale/terraform-provider-exoscale/commit/3c2a20ba9090f66ae338fdee9bed3948739baa6c

pkg/resources/nlb/models.go
pierre-emmanuelJ: 3c2a20ba

pkg/resources/nlb/resource.go
pierre-emmanuelJ: Fixed. `Create` and `Update` now write only the computed attributes (`applyNLBComputed` / `applyServiceComputed`); the full refresh (`applyNLB` / `applyService`) is used by `Read` alone, with the reasoning in a comment on the helper.

Worth noting this also removed a workaround: I had been resetting `description` back to the planned value after the refresh specifically to dodge that error. Once Create/Update stop overwriting configured attributes, the special case disappears on its own.

pierre-emmanuelJ: 3c2a20ba

pkg/resources/nlb/resource.go
pierre-emmanuelJ: You are right on both counts, and I have dropped that reasoning entirely. Both resources now send only what changed.

Two real constraints turned up while testing it, neither of which is "partial updates are rejected":

**The load-balancer endpoint rejects an *empty* payload, not a partial one.** Every field is `omitempty`, so the step that only clears `labels` serialised to `{}` and got a 5xx. `update` is now set only when a field will actually be serialised — clearing labels goes through `ResetLoadBalancerField`, so it is not an update at all.

**The service endpoint does require `protocol`.** This one is independent of body size; it reproduces with a perfectly non-empty body:

```
PUT /v2/load-balancer/{id}/service/{sid}
{"healthcheck":{"interval":10,"mode":"tcp","port":8443,"retries":1,"timeout":5}}

400 Invalid value `:HTTP/1.1` at 'protocol' (description: `Network traffic protocol`):
    should be one of :tcp, :udp
```

The `:HTTP/1.1` is the server misreporting an absent field — it is not something the provider sends. So `protocol` and `strategy` are always included in the service payload (both always hold a value, from config or the schema default); everything else is conditional. If that is a known API bug worth fixing server-side, happy to drop the workaround and link an issue instead.

pkg/resources/nlb/resource.go
pierre-emmanuelJ: 3c2a20ba

pierre-emmanuelJ:
Pushed `3c2a20ba` addressing all six review comments — thanks, these were good catches.

Verified after the changes: `TestNLB` **PASS** (134s, all 5 steps including the merged data-source coverage), plus build / vet / gofmt / staticcheck / unit tests / docs-idempotency locally.

One correction I owe the PR description: it claimed *"the API rejects a partial update"*. That was wrong. The load-balancer endpoint accepts partial payloads and only rejects an **empty** one; separately, the service endpoint requires `protocol`. Details in the thread on `resource.go`. I will fix the PR body before this comes off hold.

pkg/resources/nlb/resource.go
pierre-emmanuelJ: Checked the OpenAPI spec and tested the hypothesis against the API. Summary: it is an API/spec mismatch, and it is **not** tied to replacing the healthcheck.

`update-load-balancer-service` in `v3/generator/source.yaml` declares `name`, `description`, `protocol`, `strategy`, `port`, `target-port`, `healthcheck` with **no `required:` list at all** (same for `load-balancer-service-healthcheck`). So egoscale generating `omitempty` everywhere is faithful to the spec — the server is what disagrees.

Ran the matrix against a live service to test @predrag's theory that replacing the healthcheck is what makes attributes required:

| body sent | result |
|---|---|
| `{"description":"…"}` | 400 `protocol` |
| `{"port":81}` | 400 `protocol` |
| `{"healthcheck":{…}}` | 400 `protocol` |
| `{"healthcheck":{…},"strategy":"round-robin"}` | 400 `protocol` |
| `{"healthcheck":{…},"protocol":"tcp"}` | 200 |
| `{"healthcheck":{…},"protocol":"tcp","strategy":"round-robin"}` | 200 |

Rows 1 and 2 carry no healthcheck and still fail, so the healthcheck theory does not hold: **`protocol` is required on every service update, unconditionally.**

Two side findings:
- **`strategy` is not required** (row 5 passes without it). I had been sending it on inference, which was wrong — narrowed to `protocol` alone in `43b77ac6`.
- The `` `:HTTP/1.1` `` in the message is the server misreporting an *absent* field, not a value the client sends. Worth fixing alongside, it sends you down the wrong path.

So either the API should honour the spec and accept a partial update, or the spec should gain `required: [protocol]`. Happy to drop the workaround and link an issue once that is decided — the comment in the code says as much.

pkg/resources/nlb/resource.go
```diff
@@ -0,0 +1,468 @@
...
+		resp.Diagnostics.AddError("API returned an error while fetching the updated NLB", err.Error())
+		return
+	}
+
+	resp.Diagnostics.Append(applyNLBComputed(&plan, nlb)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```
kobajagi: This is a fragile pattern. It may work fine in this case because it is very simple resource, but it can also crash in non-obvious way. It also doesn't allow partial updates (saving terraform state between 2 API calls).

On Update you have plan struct and state struct. Normally you sync remote from plan, apply changes from plan struct to state struct and then save the state. Here you are using plan struct directly instead.

PR #596, opened 2026-09-18
Lei/add vpc dhcp options

maximilian1064:
# Description

## Checklist
(For exoscale contributors)

* [ ] Acceptance tests OK


examples/resources/exoscale_vpc/resource.tf
```diff
@@ -3,9 +3,12 @@ locals {
...
 resource "exoscale_vpc" "my_vpc" {
-  zone        = local.zone
-  name        = "my-vpc"
-  description = "My Virtual Private Cloud"
+  zone          = local.zone
+  name          = "my-vpc"
+  description   = "My Virtual Private Cloud"
+  dns_servers   = ["8.8.8.8", "1.1.1.1"]
+  ntp_servers   = ["42.42.42.42", "43.43.43.43"]
+  domain_search = ["my.domain", "their.domain"]
```
aureliar8: Question: would it make sense to have a top-level fields `dhcp_options` which then has three fields in it ?

maximilian1064: I think it's fine to leave the options at top level. I cannot think of any problems. This is also similar to the Cli where we add an optional parameter for each of the dhcp options. LMK if I missed smg.

maximilian1064:
I wasn't aware of the validator but then I saw the discussion in another PR. Is it better to write a validator instead of doing the check in the create/update functions? @exoscale/giza @sauterp

pkg/resources/vpc/datasource.go
```diff
@@ -175,13 +197,17 @@ func (d *DataSource) Read(ctx context.Context, req datasource.ReadRequest, resp
...
-		Name:        types.StringValue(vpc.Name),
-		Zone:        state.Zone,
-		Description: types.StringValue(vpc.Description),
-		Default:     types.BoolValue(vpc.Default != nil && *vpc.Default),
-		Timeouts:    state.Timeouts,
+		ID:           types.StringValue(vpc.ID.String()),
+		Name:         types.StringValue(vpc.Name),
+		Zone:         state.Zone,
+		Description:  types.StringValue(vpc.Description),
+		Default:      types.BoolValue(vpc.Default != nil && *vpc.Default),
```
aureliar8: nit and i know it doesn't come from this PR but the default field can't be nil, thus you can simply dereference here

examples/resources/exoscale_vpc/resource.tf
sauterp: Each of these options can be set independently of the others so I think it's fine if they aren't grouped. If we want to make it very clear to the user that these are DHCP options, then a nested block makes sense. But the attribute descriptions already state that. I don't have a preference, so I'd keep it the way it's now :)

examples/resources/exoscale_vpc/resource.tf
```diff
@@ -3,9 +3,12 @@ locals {
...
 
 resource "exoscale_vpc" "my_vpc" {
-  zone        = local.zone
-  name        = "my-vpc"
-  description = "My Virtual Private Cloud"
+  zone          = local.zone
+  name          = "my-vpc"
+  description   = "My Virtual Private Cloud"
+  dns_servers   = ["8.8.8.8", "1.1.1.1"]
+  ntp_servers   = ["42.42.42.42", "43.43.43.43"]
```
sauterp: I noticed while testing this example that if I set an empty list `[]`, I get permanent drift, meaning I can run `terraform apply` over and over and I get a non-empty plan. Unfortunately the terraform framework that we use distinguishes between empty and null Lists. You can fix it with something like this:
https://github.com/exoscale/terraform-provider-exoscale/pull/585/changes#diff-07a653aa93ddabc2491bca9d612bdde4a28b4159e706b6ebbc2d7d47c2c13a9dR380-R389

sauterp (changes requested):

sauterp:
> I wasn't aware of the validator but then I saw the discussion in another PR. Is it better to write a validator instead of doing the check in the create/update functions? @exoscale/giza @sauterp

IMO you can remove the validation in terraform if we have validation in our API. As was pointed out in the other discussion, it's better to have validation in only one place and centrally if possible.

maximilian1064:
> > I wasn't aware of the validator but then I saw the discussion in another PR. Is it better to write a validator instead of doing the check in the create/update functions? @exoscale/giza @sauterp
> 
> IMO you can remove the validation in terraform if we have validation in our API. As was pointed out in the other discussion, it's better to have validation in only one place and centrally if possible.

Ack. The difference here is that public API already specifies dns/ntp servers as IPv4, while subnet/cidr-block is specified as string. So I don't expect we'll change the type. And if we ever do, we'll need to change the resource code anyway as the egoscale type will change.

Furthermore, in egoscale dns/ntp servers have type net.IP, so it's natural to do some validation like

```go
ip := net.ParseIP(ipStr)
if ip != nil && ip.To4() != nil {
```
when parsing the IPs.

So I would say we can do the validation here.

sauterp:
Yes in this case, since egoscale expects a `net.IP`, we don't have a choice but to parse/validate.

examples/resources/exoscale_vpc/resource.tf
maximilian1064: thx for catching this. [7b55f36](https://github.com/exoscale/terraform-provider-exoscale/pull/596/commits/7b55f36875ae7e68ed67b0e54ce20971a9fc090c)

sauterp (approved):
Thanks!

docs/resources/vpc.md
```diff
@@ -21,9 +21,12 @@ locals {
 }
 
 resource "exoscale_vpc" "my_vpc" {
-  zone        = local.zone
-  name        = "my-vpc"
-  description = "My Virtual Private Cloud"
+  zone          = local.zone
+  name          = "my-vpc"
+  description   = "My Virtual Private Cloud"
+  dns_servers   = ["8.8.8.8", "1.1.1.1"]
```
nerzhul: we cannot put IPv6 DNS ? could be nice if i run rtadvd on my VPC (as we don't provide v6 natively) or if i want to ensure IPv6 are using the external v6 address of my machine ?

What about option conflict there with the DHCP on regular interface outside VPC interfaces ?

maximilian1064: VPC MVP is Ipv4 only, so what we have here are only dhcpv4 options (I probably should make it clear in the doc). We'll add IPv6 support and the relevant dhcp options when we have VPC GA.

It's up to the customer to handle the conflicts (see [this](https://app.shortcut.com/exoscale/story/194226/vpc-allow-customer-to-set-dhcp-options-like-they-do-on-privnet#activity-196456) for potential conflicts for each of these options). This is the same in private networks (https://community.exoscale.com/product/networking/private-network/how-to/dhcp-options/).

aureliar8: >  could be nice if i run rtadvd on my VPC (as we don't provide v6 natively)

For the MVP we actively drop all ipv6 packet that leaves instance on a VPC interface, so you won't be able to do it anyway

pkg/resources/vpc/datasource.go
kobajagi: nil pointer dereference is a panic and API bugs slip in occasionally  :wink: 
That being said, there exists a `types.BoolPointerValue` for this case.

aureliar8: Fair enough, we can add it just in case

docs/resources/vpc.md
nerzhul: ok perfect :) thanks for your feedback

pierre-emmanuelJ (approved):

docs/resources/vpc.md
```diff
@@ -21,9 +21,12 @@ locals {
...
+  ntp_servers   = ["42.42.42.42", "43.43.43.43"]
+  domain_search = ["my.domain", "their.domain"]
```
aureliar8: Maybe I'd suggest to not put those option in this doc. 

In other pages this is where we showcase the simple example, not the most complex one labelhttps://registry.terraform.io/providers/exoscale/exoscale/latest/docs/resources/compute_instance

The fields are still documented in the list of fields, but if a customer want a quick copy paste they don't need it

maximilian1064: [ac4b16e](https://github.com/exoscale/terraform-provider-exoscale/pull/596/commits/ac4b16e21c4a178ccc0afdb070161a6f20f81767)

